PR #72, opened 2022-05-30
Update 0-cluster-autoscaler.yaml

shaypi:
This will fix the cluster-autoscaler on eks version v1.22.2

### DIFF
--- a/lessons/070/k8s/0-cluster-autoscaler.yaml
+++ b/lessons/070/k8s/0-cluster-autoscaler.yaml
@@ -37,6 +37,7 @@ rules:
     verbs: ["watch", "list", "get", "update"]
   - apiGroups: [""]
     resources:
+      - "namespaces"
       - "pods"
       - "services"
       - "replicationcontrollers"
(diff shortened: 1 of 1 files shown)


antonputra:
> This will fix the cluster-autoscaler on eks version v1.22.2

Thanks for PR, what is broken exactly?

shaypi:
> 
When I deployed EKS version 1.22 there was an issue with the ClusterRole(I got some error) and I managed to resolve it by using the fix in the PR.

antonputra:
> > 
> 
> When I deployed EKS version 1.22 there was an issue with the ClusterRole(I got some error) and I managed to resolve it by using the fix in the PR.

Thanks, will verify and merge

antonputra:
Based on the documentation, minor version (major.minor.patch) of cluster autoscaler should match the Kubernetes minor version. That said, new cluster autoscalers may have different requirements (different RBAC permissions maybe needed). I'll close this PR since it relevant only to new 1.22 versions.

PR #100, opened 2022-09-27
Lesson 128 Improvments.

Bajix:
This PR makes things a bit more idiomatic. Perhaps the profile optimization changes will affect performance

Also try:
```
cargo lambda build --release --arm64
```


antonputra:
Thanks for the PR!

PR #194, opened 2023-09-18
Update main.yaml

Ayoub-HEDFI:
avoid Errors on new AWS Ubuntu AMIs, need to update_cache like sudo apt update before trying to install Nginx, if not it won't the find its package in apt,.... Thanks btw

### DIFF
--- a/lessons/014/roles/nginx/tasks/main.yaml
+++ b/lessons/014/roles/nginx/tasks/main.yaml
@@ -1,8 +1,14 @@
 ---
+- name: Update package cache
+  become: yes
+  apt:
+    update_cache: yes
+
 - name: Ensure Nginx is at the latest version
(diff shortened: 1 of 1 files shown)


antonputra:
@Ayoub-HEDFI thanks for the PR, but
1. "Become: yes" is already set on the playbook here - https://github.com/antonputra/tutorials/blob/603202344f0a1c1bb07cad3f1fb370db9311dbbd/lessons/014/nginx.yaml#L5
1. Also I would prefer to include "update_cache: yes" on the existing task  rather then create a new one - https://github.com/antonputra/tutorials/blob/603202344f0a1c1bb07cad3f1fb370db9311dbbd/lessons/014/roles/nginx/tasks/main.yaml#L8

PR #246, opened 2024-08-09
feat: use argocd applicationsets

thatmlopsguy:
Use argocd applicationsets and organize the argocd application in a project.

after running this command
kubectl apply -f lessons/200/argocd/root-app.yaml

two applications will be generated for each environment inside the `tutorial-app-of-apps` project.


antonputra:
Thanks for the PR!

PR #249, opened 2024-08-23
Performance improvements of the cs-app

Odonno:
* [Avoid use of string.Format](https://github.com/antonputra/tutorials/commit/6227f409beefbf7789a611798728f8f317d610d4)

* [Use built-in configuration binding](https://github.com/antonputra/tutorials/commit/169d9e04f692f7568b535e50965a94dd2afb17a8)

* [Use sealed class/readonly struct](https://github.com/antonputra/tutorials/commit/2dfc1e6b3c88e33cb593cdd990bb6de231547b74)
* [Dispose AmazonS3Client](https://github.com/antonputra/tutorials/commit/0e3db73e44058e638a0efdbab5dee8a944a6fcd1)

* [Make web api aot compatible](https://github.com/antonputra/tutorials/commit/8e74f5c80e7008d9ba7dbb0582e502fffef76837)

It almost reduces by 3x the memory usage locally. I did not perceive any CPU gain but I suppose there are some when publishing in release mode.

AOT should also drastically reduce the output image size.


lessons/202/cs-app/Program.cs
```diff
@@ -1,97 +1,111 @@
...
+    ];
+
+    return Results.Ok(devices);
+});
+
+// Create endpoint that uoloades image to S3 and writes metadate to Postgres
+app.MapGet("/api/images", async () =>
+{
+    // Generate a new image.
+    var image = new Image($"cs-thumbnail-{counter}.png");
```
alisonmashni, line 71: Should counter be used with `Interlocked.Increment(ref counter)` instead of reading it and later using counter++?

lessons/202/cs-app/Program.cs
```diff
@@ -1,97 +1,111 @@
...
+
+    // Get the current time to record the duration of the S3 request.
+    var s3Stopwatch = Stopwatch.StartNew();
```
alisonmashni, line 74: It would probably be more efficient to replace the Stopwatch instances with `Stopwatch.GetTimestamp()`  and `Stopwatch.GetElapsedTime(start)`

lessons/202/cs-app/Program.cs
alisonmashni: Apparently the Go App is not handling resource concurrency there. If it's just to measure raw performance then it may be better to run without Interlocked.Increment after all. 

image

I'm sorry for suggesting it before checking how the Go App was working before.

Odonno: Hey, can't bite to make it right. Can't see why we can't fix the Go app too, but I am not sure how. Or better, we could send the incremented `id` from the client via a route param.

antonputra (changes requested):
thanks for the improvement! I want to keep original "cs-app" app for the reference. can you copy paste it to "cs-app-v2" under the same "202" lesson and I'll merge it

Odonno:
Done. I named the folder `cs-app-aot`. I hope it's ok for you @antonputra

Odonno:
Note: the next version of the AWS SDK for .NET will bring a massive cut of the package size (2MB to 900kb, and that's just without trimming). It will also improve performance and provide AOT support.

A preview version is available via NuGet but I am not sure that would work. If you have time to try it, that can be interesting to do the migration.

Blog post for reference: https://aws.amazon.com/fr/blogs/developer/preview-1-of-aws-sdk-for-net-v4/

sergeyshaykhullin:
At 28 April V4 AWS SDK is GA

https://aws.amazon.com/blogs/developer/general-availability-of-aws-sdk-for-net-v4-0/

PR #251, opened 2024-08-31
Try to improve Rust performance for Lesson 203

devashishdxt:
I've tried to improve the performance of Rust code for lesson 203. Ideally, I'd use `axum` (for http server) and `metrics-exporter-prometheus` for metrics but I haven't changed that as this comparison is between `actix-web` and Go framework.

### DIFF
--- a/lessons/203/rust-app/Cargo.lock
+++ b/lessons/203/rust-app/Cargo.lock
@@ -1889,6 +1889,7 @@ dependencies = [
  "serde",
  "serde_derive",
  "serde_json",
+ "tokio",
  "tokio-postgres",
  "toml",
  "uuid",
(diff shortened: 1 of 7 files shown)


antonputra:
Thanks for the PR @devashishdxt , i'd like to keep original code as is just for the future references, would you mind to create something like rust-app-v2 or improved folder in this 203 lesson and I'll merge and test the code?

antonputra (changes requested):
Thanks for the PR @devashishdxt , i'd like to keep original code as is just for the future references, would you mind to create something like rust-app-v2 or improved folder in this 203 lesson and I'll merge and test the code?

devashishdxt:
> Thanks for the PR @devashishdxt , i'd like to keep original code as is just for the future references, would you mind to create something like rust-app-v2 or improved folder in this 203 lesson and I'll merge and test the code?

Ok. Will create another PR.

PR #252, opened 2024-08-31
Lesson 203: Remove String to Uuid conversion to closely resemble go

divitngoc:
When hitting devices endpoint, it's a bit unfair as Rust has to do extra computation compared to Go from String to Uuid and then back Uuid to String when serializing.


antonputra (approved):
true, good catch

PR #256, opened 2024-09-07
A few Go improvements

wltechblog:
1. The Go version was using a dynamic table name supplied as a function argument, which is not present in the Rust version and which requires an extra fmt.Sprintf to pre-prepare the query. I made it static to match the Rust version.
2. Made the image save() function a method on the Image type, which is more idiomatic Go, probably no performance difference but it's cleaner
3. Removed an unnecessary variable creation in NewImage() and assign the CreatedAt direction from  time.Now() 
4. Since the components of Key are all strings,  we should use string concatenation instead of fmt.Sprintf to create the Key.

### DIFF
--- a/lessons/205/go-app/images.go
+++ b/lessons/205/go-app/images.go
@@ -45,15 +45,12 @@ func NewImage() *Image {
 }
 
 // Save inserts a newly generated image into the Postgres database.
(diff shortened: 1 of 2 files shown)


Adembc (approved):
LGTM

PR #257, opened 2024-09-07
Use prepare_cached instead of prepare

joshka:
Aligns with the go implementation.

### DIFF
--- a/lessons/205/rust-app/src/image.rs
+++ b/lessons/205/rust-app/src/image.rs
@@ -72,10 +72,16 @@ impl Image {
             .context("failed to get database connection")?;
 
         // Prepare the SQL query.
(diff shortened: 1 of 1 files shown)


devashishdxt (commented):
If you don’t pass prepared statement, tokio-postres will internally call prepare function on the string.

joshka:
Damn - you're right. I missed that.
Reading through the go implementation more, its postgres connection has an internal cache for prepared statements, which it uses when running any plain sql query https://pkg.go.dev/github.com/jackc/pgx#Conn.Prepare

joshka:
Switched to calling prepare_cached() instead. This seems like it's probably closer to the go implementation.

PR #258, opened 2024-09-07
remove cow and add concurrent upload/save

BJSummerfield:
**Use static str for devices**

[Cow](https://doc.rust-lang.org/std/borrow/enum.Cow.html) adds unnecessary overhead.  We are creating and returning static strings with devices, never using the `Owned` enum variant.  With Cow, each device creates an enum that holds references to an  `Owned` memory location and/or a `Borrowed` memory location.  However, this program only uses `Borrowed` path.

Removing the overhead creates a small but measurable improvement.

Local stress test using Cow running `wrk -t12 -c400 -d30s http://localhost:8080/api/devices`
Requests/sec: 
171901.99
172399.36
172104.08

Using static strings
Requests/sec: 
173199.04
173038.07
172948.25

**Image save and upload concurrently**

No need to wait for image upload to happen before save (unless we want to).  Run them both at the same time with `tokio::join`

### DIFF
--- a/lessons/203/rust-app-v2/src/device.rs
+++ b/lessons/203/rust-app-v2/src/device.rs
@@ -1,10 +1,8 @@
-use std::borrow::Cow;
-
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Device<'a> {
-    pub uuid: Cow<'a, str>,
-    pub mac: Cow<'a, str>,
-    pub firmware: Cow<'a, str>,
+    pub uuid: &'a str,
+    pub mac: &'a str,
+    pub firmware: &'a str,
 }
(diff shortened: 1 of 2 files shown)


antonputra:
Thanks @BJSummerfield for the PR, let me test it and I'll merge

antonputra:
What do you think? @devashishdxt

devashishdxt:
Using `&'a str` is definitely better than using `Cow`. I also thought of this but decided not to do that because in a real world application, you rarely return static strings which are known at compile time.

For concurrently saving and uploading, you'll also have to make changes in go application so that go executes both in parallel. Doing it concurrently will be faster in both the languages.

antonputra:
Thanks, i'll update go code before the next test!

antonputra (approved):

PR #259, opened 2024-09-07
Add reactive java app version

ericsouza:
hey @antonputra , this is my pr of the reactive version of the java app for the lesson 201, which is a comparison between quarkus (java) and golang (fiber). I think I covered most of the blocking part. There is only the AtomicInteger usage, which is indeed thread safe but replacing with UUID can increase perfomance. I do still want to try the vertx Counter which is a properly way to have an atomic counter in a vertx/reactive environment.

for now, I think it can be already used

btw, Is there any problem changing the key of image file from using a counter to use an uuid as follow?
```java
String imgKey = "java-thumbnail-" + UUID.randomUUID() + ".png";
Image img = new Image(imgKey);
```

thanks for your working with all these benchmarks!!


antonputra:
Thanks @ericsouza for the PR, let me test it and i'll merge

lessons/201/java-app-reactive/pom.xml
```diff
@@ -0,0 +1,152 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>com.antonputra</groupId>
+    <artifactId>java-app</artifactId>
+    <version>0.1.0</version>
+
+    <properties>
+        <compiler-plugin.version>3.13.0</compiler-plugin.version>
```
begui, line 9: Consider a 'quarkus update' . I want to say I remember reading about some improvements to jackson in 3.14+

antonputra: thanks for the tip, I'll update quarkus version before next test

antonputra (approved):
thanks, i'll update version before next test

lessons/201/java-app-reactive/pom.xml
ericsouza: to get the hackson improvements from quarkus 3.14 there is also the need to seed the property below

```
quarkus.rest.jackson.optimization.enable-reflection-free-serializers=true
```

ericsouza:
> Thanks @ericsouza for the PR, let me test it and i'll merge

Thanks!

If you don't see a big difference, please consider removing the counter used by the upload method and replacing with an uuid string

franz1981:
Thanks @ericsouza for this PR; any update on this version @antonputra ?
I have the suspect that if the bottleneck in the tests is nother the application instance, using one language/framework or others won't matter that much, unless the framework is doing something terribly wrong.

franz1981:
I have noticed a couple of possible improvements (assuming the cost to not be on the external services):
- https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/http-configuration-netty.html could use Netty, and since the vertx event loops could install the AWS Netty client on it, it should reduce the number of thread handoffs
- https://github.com/eclipse-vertx/vert.x/issues/5665 reading files like that is performing useless bug allocations
- https://github.com/eclipse-vertx/vert.x/blob/4.5.16/src/main/java/io/vertx/core/buffer/impl/BufferImpl.java#L216 using this method allocate another copy of the same data 
- https://github.com/aws/aws-sdk-java-v2/blob/master/core%2Fsdk-core%2Fsrc%2Fmain%2Fjava%2Fsoftware%2Famazon%2Fawssdk%2Fcore%2Fasync%2FAsyncRequestBody.java#L206 create another copy again

PR #263, opened 2024-09-09
Preserialize Devices as &str on lesson 206

JeremyMoeglich:
Currently every request has to allocate and serialize Devices before starting to actually process the request.

This PR preserializes the Devices into a static string that is compiled into the binary. Here are my local benchmarks. The benchmark involved sending as many requests as possible to each of the servers over 10 seconds counting the amount.

debug stands for the `debug = true` option in the cargo toml aswell as running a profiler. nodebug stands for debug = false and no debugger

Actix:
debug before, Total successful: 3065795
nodebug before, Total successful: 3311823
debug after, Total successful: 3201285 + 4.42%
nodebug after, Total successful: 3423525 + 3.37%

Axum:
debug before, Total successful: 3016059
nodebug before, Total successful: 3112641
debug after, Total successful: 3320078 + 10.08%
nodebug after, Total successful: 3444513 + 10.65%

Rocket:
debug before, Total successful: 2370935
nodebug before, Total successful: 2459899
debug after, Total successful: 2464806 + 3.96%
nodebug after, Total successful: 2506760 + 1.91%

The differences are likely lower in a network setting since I tested everything locally, but not doing something should always be faster


antonputra (changes requested):
Thanks @JeremyMoeglich for the PR. I understand that this improves performance, but it makes the application static and not practical at all. If you could keep the original code as it is, and perhaps create an actix-app-static or actix-app-v2 version, I would be willing to merge it after testing.

antonputra (approved):
you didn't have to copy tests but it's fine

PR #264, opened 2024-09-10
[205] Increasing socket count

alshdavid:
The default `TcpListener` in Rust has a very conservative max socket count (lower than Go). Above that threshold, Rust will start dropping socket connections (even if the hardware has the capability to process more).

Depending on the number of concurrent connections configured in the stress testing tool, given you still had CPU and RAM left, it's possible that you simply hit the limit for socket connections and the Rust server started dropping new connections.

image

This PR increases the max socket count to the max amount possible which should let the Rust server dig a little deeper.

### DIFF
--- a/lessons/205/rust-app/src/main.rs
+++ b/lessons/205/rust-app/src/main.rs
@@ -5,7 +5,12 @@ mod metrics;
 mod routes;
 mod state;
 
-use std::{future::ready, io, net::{SocketAddr, ToSocketAddrs}, time::Duration};
+use std::{
+    future::ready,
+    io,
+    net::{SocketAddr, ToSocketAddrs},
+    time::Duration,
+};
 
 use axum::{http::StatusCode, routing::get, Router};
 use metrics::setup_metrics_recorder;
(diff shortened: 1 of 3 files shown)


lessons/205/rust-app/src/main.rs
```diff
@@ -54,16 +55,36 @@ async fn main() {
...
+    SocketAddr::V4(_) => Socket::new(Domain::IPV4, Type::STREAM, None)?,
+    SocketAddr::V6(_) => Socket::new(Domain::IPV6, Type::STREAM, None)?,
+  };
+
+  listener.set_nonblocking(true)?;
+  listener.set_nodelay(true)?;
+  listener.set_reuse_address(true)?;
+  listener.set_linger(Some(Duration::from_secs(0)))?;
+  listener.bind(&addr.into())?;
+  listener.listen(i32::MAX)?;
```
alshdavid, line 85: Increasing the max socket count

antonputra:
Nope, it didn't help. I keep hitting that magic number of 27k RPS, no matter how much memory or CPU I allocate. I'll test on VMs to avoid an additional layer of abstraction. But it hasn't been related to socket count so far.

Screenshot 2024-09-11 at 3 36 10 PM

PR #266, opened 2024-09-12
Go cleanups for tutorial 205

cookieo9:
Laying a little groundwork to make the code a bit easier to work with. The actual performance impact (of the changes that have any) is minimal and lost in the noise caused by allocations, the GC, and network traffic.

## Main changes

 * Using io.WriteString to clean up writing strings (esp. string constants)
 * Better use of contexts:
   * Piping them through to allow re-use
   * Creating appropriate contexts for init-time and per-request
 * Modularizing the parts of server initialization to ease testing of base behaviour
 * Adding tests and benchmarks for the "easy stuff"
 * Create a runtime profile from the benchmarks to use with PGO

## A note on PGO

PGO (Profile Guided Optimization) is now available, but not enabled by default. This was for a few reasons:
 1. This changeset is about cleanups not performance.
 2. The benefits are easy to lose in the noise, especially while other things (e.g. allocations, garbage collection, networking) have a more noticable impact.
 3. A more representative profile (i.e. from a real server) would be better than generating one from synthetic benchmarks.

I'll include a change to enable it by default if requested, or should I find a case where it makes a noticable difference (e.g. after some allocations are optimized away).

PGO can improve optimisations in a few ways:
 - functions/methods that are called a lot are given a higher inlining threshold allowing for deeper inlining
 - interface types can be "devirtualized" should the profile show that a given implementation of an interface is commonly used at a callsite.
 - Both of the above allows for better escape analysis, and flattens the call stack meaning the compiler is more likely to know it's safe to leave an object on the stack, and has more opportunities to do so, reducing allocations and GC work.

For more details: https://go.dev/doc/pgo

### Using PGO

There are 2 easy ways to build with PGO, as although there is a flag available to the compiler `-pgo <file>` to manually specify a profile to use, simply having a CPU profile in the file `default.pgo` in the directory of the main package will have it used as part of the build of that binary. If the profile contains data about other packages (e.g. the runtime, http server, Garbage collector, and 3rd party libraries) they will be rebuilt using the info from the profile too.

#### Create a profile

I created a task for `go generate` that runs the benchmarks and outputs the result as `default.pgo`. This file is automatically used if present, and will even apply automatically to the container builds because the `Dockerfile` copies the entire contents of the source directory. Updating this file with each code change is an option, as well as replacing it with a profile from a real server.

To use PGO, simply run `go generate` before building the server. To stop, delete the generated file (`default.pgo`). Be sure to run `go generate` after most code changes otherwise the profile will become stale.

#### Run `go generate` as part of building the container

This is a nice option as it ensures that you don't forget to update the profile every time the code changes. There are downsides:
 * It slows down the build by about 30s (likely not an issue)
 * It makes it harder / complicated to introduce a server-based profile because the `Dockerfile` will be generating one unconditionally
 * "Turning it off" to make comparisons require's updating the `Dockerfile` instead of simply `rm default.pgo`

To do this basically add this above the line to build the server:
```
RUN go generate
```

### DIFF
--- a/lessons/205/go-app/images.go
+++ b/lessons/205/go-app/images.go
@@ -45,12 +45,12 @@ func NewImage() *Image {
 }
 
 // Save inserts a newly generated image into the Postgres database.
-func (i *Image) save(dbpool *pgxpool.Pool, m *metrics) error {
+func (i *Image) save(ctx context.Context, dbpool *pgxpool.Pool, m *metrics) error {
 	// Get the current time to record the duration of the request.
 	now := time.Now()
 
 	// Execute the query to create a new image record (pgx automatically prepares and caches statements by default).
-	_, err := dbpool.Exec(context.Background(), "INSERT INTO `go_image` VALUES ($1, $2, $3)", i.ImageUUID, i.Key, i.CreatedAt)
+	_, err := dbpool.Exec(ctx, "INSERT INTO `go_image` VALUES ($1, $2, $3)", i.ImageUUID, i.Key, i.CreatedAt)
 	if err != nil {
 		return fmt.Errorf("dbpool.Exec failed: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


antonputra (approved):
Thanks, I'll go over it and test before the next benchmark

PR #271, opened 2024-09-16
Lesson 207 - add pure Zig implementation

zigster64:
Keeping the ZAP version intact, I have added a 3rd implementation here that does the same /healthz and /api/devices endpoints ... but using a pure zig lib instead

This creates an additional deployment 'zig-app' to go next to 'zap-app'. Hope that makes sense

This uses async IO (ie - old school worker pool of threads + non-blocking IO)

I have fine-tuned the server config in this case to reduce memory usage ... this should have no effect on performance, it's just to reduce runtime memory consumption.  You should in theory be able to run this in a pod with only 4 MB memory allocation !! 

Running this on local (Mac M2), im seeing around the same perf on Zig vs Rust.

Best Rust run is around 185k r/s on my machine
Best Zig run is around 198k r/s on my machine

Zig marginally better across the board, but more scatter in the results.  Much of a Muchness

A Rust expert could probably tune the Actix build to get closer to the Zig peak figure ?  

Memory usage and binary footprint significantly better than Rust 

Im not seeing any significant improvements in the Zap version by applying different tweaks. I am not familiar with Zap or facil so much

-----------

Later on, will followup with an extension to lesson 205, to do some Postgres and S3 (minio) updates inline with the Go version

Enjoy

### DIFF
--- a/lessons/207/http.zig/Dockerfile
+++ b/lessons/207/http.zig/Dockerfile
@@ -1,0 +1,25 @@
+FROM debian:12.7 AS build
+
+ARG ZIG_VER=0.14.0-dev.1511+54b668f8a
+
+RUN apt-get update && apt-get install -y curl xz-utils
+
+RUN curl https://ziglang.org/builds/zig-linux-$(uname -m)-${ZIG_VER}.tar.xz -o zig-linux.tar.xz && \
+    tar xf zig-linux.tar.xz && \
+    mv zig-linux-$(uname -m)-${ZIG_VER}/ /opt/zig
+
+WORKDIR /app
+
+COPY . .
+
+# Using ReleaseSmall instead of ReleaseFast - because there is no CPU intensive work
+# in this app, so ReleaseSmall will give a smaller binary / better use of CPU cache
+RUN /opt/zig/zig build -Doptimize=ReleaseSmall -Dcpu=baseline
+
+FROM debian:12.6-slim
+
+RUN apt-get update && apt-get install -y ca-certificates
+
+COPY --from=build /app/zig-out/bin/http.zig /server
+
+CMD ["/server"]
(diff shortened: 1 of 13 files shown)


lessons/207/http.zig/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+    tar xf zig-linux.tar.xz && \
+    mv zig-linux-$(uname -m)-${ZIG_VER}/ /opt/zig
+
+WORKDIR /app
+
+COPY . .
+
+# Using ReleaseSmall instead of ReleaseFast - because there is no CPU intensive work
+# in this app, so ReleaseSmall will give a smaller binary / better use of CPU cache
+RUN /opt/zig/zig build -Doptimize=ReleaseSmall -Dcpu=baseline
```
zigster64, line 17: or try ReleaseFast anyway on your setup ?

lessons/207/http.zig/src/main.zig
```diff
@@ -0,0 +1,57 @@
...
+        },
+    }, &app);
+    var router = server.router(.{});
+    router.get("/healthz", App.healthz, .{});
+    router.get("/api/devices", App.getDevices, .{});
+
+    std.debug.print("http.zig listening on http://0.0.0.0:{d}\n", .{port});
+    try server.listen();
+}
+
```
zigster64, line 44: Note - App is a struct, and we created an instance of this in main, and wired our server up to it

All requests then get a handle to the App instance. You can then do things like have a DB object in the App instance, and other state information like that - so all requests can easily that App state

You can also have multiple different Objects instantiated, and wire different endpoints to different instances

Its pretty similar to Go from that regard ... and I think ... reasonably easy to follow for anyone familiar with Go

The middleware API is similar to Go as well

antonputra (approved):
I appreciate your help. I'll release a new video with the results in a day or two.

websitegardener:
@zigster64 could you do a short screen recording of your implementation? The new test that @antonputra did on his youtube video shows the opposite

PR #274, opened 2024-09-17
lesson 208: set 2 workers is similar to that in rust

hessenar:
I am not a strong expert in zig, but according to the code, actix uses 2 workers (and it seems like 2 threads), I suggest making similar logic for the zig code

### DIFF
--- a/lessons/208/http.zig/src/main.zig
+++ b/lessons/208/http.zig/src/main.zig
@@ -18,11 +18,12 @@ pub fn main() !void {
         .address = "0.0.0.0",
         .port = port,
         .workers = .{
+            .count = 2,
             .large_buffer_count = 1,
             .large_buffer_size = 1024,
         },
         .thread_pool = .{
-            .count = 1,
+            .count = 2,
             .buffer_size = 1024,
         },
         .request = .{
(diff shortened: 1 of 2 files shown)


antonputra:
Actually i verified this one, it see only 1 cpu inside the container and uses 1 worker not 2.

"by default this number is equal to the number of physical CPUs"

antonputra:
@hessenar I even tried running this app on a standalone VM, 'm7a.large,' with 2 vCPUs. The problem is that the 2 vCPUs correspond to 1 physical processor with two cores, and the app only sees 1 physical processor, so it starts 1 worker instead of two.

https://browser.geekbench.com/v5/cpu/22475411

hessenar:
in the actix documentation
https://actix.rs/docs/server/#multi-threading 
and in code you set 2
https://github.com/antonputra/tutorials/blob/c4eb063422957ddacde6076f2d41c8ebd19dddf7/lessons/208/actix-app/src/main.rs#L12
the presence of only one core (judging by the description, including here) is not an obstacle to creation a separate application instance to handle requests

antonputra:
@hessenar that's try i probably forgot to remove it after testing

antonputra (approved):

PR #276, opened 2024-09-17
cache prepared stm

dsimushkin:
cache the prepared stmt on the client conn. gives much better performance (esp. in case of benching a single request).


antonputra:
@dsimushkin pls take a look at the discussion in this PR - https://github.com/antonputra/tutorials/pull/257

antonputra (approved):

dsimushkin:
Yep, that's it. Thanks for your time.

PR #277, opened 2024-09-18
increase worker max_conn

karlseguin:
Not sure if this is why httpz flopped in the end.  If this helps, it's obviously something I need to handle better automatically without needing it tweaked. But if you could try with this small change, I'd appreciate it :)

### DIFF
--- a/lessons/207/http.zig/src/main.zig
+++ b/lessons/207/http.zig/src/main.zig
@@ -18,6 +18,7 @@ pub fn main() !void {
         .address = "0.0.0.0",
         .port = port,
         .workers = .{
+            .max_conn = 4096,
             .large_buffer_count = 1,
             .large_buffer_size = 1024,
         },


zigster64:
Thx Karl !

Apologies for jumping the gun on this one. I was hoping to try a few configs in the actual environ before "going live" with this, but ... here we are :)

I believe we are constrained to 2 vCPUs on the target machine - not sure

In local testing I currently have 8k for the max conns, 2 request workers, and thread pool = 2.  That seems to give peak performance on my M2, but I have no idea how the target env will react to that yet

Downside of boosting max conns is to see the memory usage go up much higher on boot (but still remain stable)

Very very pleased to see the memory usage graph be a flat line though - I remember that an awful lot of hard work into making that happen :). Looks like it works in a big environment

zigster64:
Also - you want to apply this change to lesson/208 instead of 207 I think

(I think that's how Anton is working this)

karlseguin:
> Downside of boosting max conns
Set `min_conn`, else it takes `max_conn`'s value.

i'll close this in favor or what you're working on

PR #279, opened 2024-09-19
Go parallel getImage handler, better errors, and fix DB

cookieo9:
To match what Rust did in #258 (making save-image parallel) do something similar for Go. Also, insert a transaction such that the DB will never be updated unless the S3 upload succeeds by having a DB transaction wrap the whole process.

Fixed up error code with a helper to clean-up some returns, as well as standardize wrapping errors.

Finally, fix a bug that broke the DB submit code in PR #256 when it used the wrong quoting style for postgres databases. Although I am having trouble configuring Minio properly to do a full test, I was able to see this failure and test that my change worked (when S3 errors were ignored).

### DIFF
--- a/lessons/205/go-app/go.mod
+++ b/lessons/205/go-app/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/jackc/pgx/v5 v5.6.0
 	github.com/prometheus/client_golang v1.20.2
 	go.uber.org/automaxprocs v1.5.3
+	golang.org/x/sync v0.8.0
 	gopkg.in/yaml.v2 v2.4.0
 )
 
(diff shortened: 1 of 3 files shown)


antonputra:
@cookieo9 you can use docker compose to run it locally, postgres and monio

```
---
services:
  minio:
    image: minio/minio:RELEASE.2024-09-13T20-26-02Z
    command: server /data --console-address ":9001"
    ports:
      - 9000:9000
      - 9001:9001
    networks:
      - private
    volumes:
      - minio-data:/data

  postgres:
    image: postgres:16.4
    ports:
      - 5432:5432
    environment:
      POSTGRES_USER: myuser
      POSTGRES_PASSWORD: devops123
      POSTGRES_DB: mydb
    volumes:
      - postgres-data:/var/lib/postgresql/data
    networks:
      - private

volumes:
  minio-data:
  postgres-data:

networks:
  private:

antonputra (approved):
Thanks, I'll merge it to match Rust, but the initial idea was to upload the image, get the 'last modified' date from AWS, and use it to save metadata about the image.

PR #280, opened 2024-09-20
208 tuning

zigster64:
A few changes to address the previous run which bombed out at 80k

Basing the assumption here that it capped out for a couple of reasons

1. Having a single worker thread - while that gave peak perf on local, it probably spun up too much CPU in the test env, and introduced Kubernetes throttling. 
2. Think the main culprit was the "max connections" config on the 2 request acceptor threads with 500 connection backlog each. This only allows a backlog of 1000 connections / accepter coroutines in practice, which may have been insufficient.
3. The handling of the overflow condition in the base library isnt particularly optimised at this point.  Its not a condition we have been running into in our test / deployment setups, so hasn't got much love yet

Addressed by 

1. Use 4 worker threads ... because why not ?  The request processing (once parsed) is dead simple, so I dont think this will have much effect. Doesnt burn up too many resources anyway, so throwing it out there
2. Use 4 request acceptor threads, each with 8k backlog.   That will take us from having max 1000 connection backlog up to 32k backlog !!   I think 4 threads should be reasonable for a 2 vCPU setup. The job of the acceptor thread is to take a new connection, read the header, parse it, then pass the request to the queue of worker threads.  Worker threads are much simpler in this case - they take the parsed request, and generate the response.  Setting it up 4:4 should be pretty balanced 
3. Using an experimental branch of the base library that takes a more defensive approach to handling the overflow condition.  WIP.  https://github.com/karlseguin/http.zig/compare/master...max_conn_handling

See build.zig.zon for where its referencing this experiment

Expectations

- Probably a bit slower on average, as its taking a "wide" approach rather than maxing out resources to get max speed.
- Expecting the memory to start growing as the app nears saturation, so it spins up more coroutines to handle the backlog
- Unsure about the CPU throttling. Haven't directly addressed this beyond adding more threads to spread the load ... so who knows ?  Only one way to find out, and I totally appreciate Anton providing this testing service 

NOTE - will followup with another patch in a minute that adds zzz as well


zigster64:
Added zzz now as well
https://github.com/mookums/zzz

It's another upcoming Zig networking framework ...  that just happens to handle being a web server as well. 

Its bleeding edge :)   zzz is tuned for use on Linux only at the moment, as it exploits io_uring.  It should perform pretty well on EKS then, given that's all linux

http.zig library is using epoll on linux to do the non-blocking IO.

Have fun !

Local tests show  krps values 

image

zigster64:
Added notes from Karl  (http.zig author)

To add.

- httpz had a very conservative max_conn default of 512. This has been increased to 8192. actix-web's limit is 25K
- httpz didn't deal with max_conn limit being hit very well, this has been improved.
- httpz doesn't automatically scale to # of cores. I generally dislike this in large part because hyperthreading cores are garbage. So, tweaking the counts for the specific machine/workload is important.

If the issue with the original test was hitting the configurable limit, then either/both the first two items should help.

There are some other tweaks in an tweaks branch which'll help minimize contention on a specific mutex, and which improves the performance of validating header values (original was taken from Go's net/http)

antonputra (approved):
@zigster64 Thanks for the PR, I'll run the hopefully next week

PR #282, opened 2024-09-21
pgx defaults to 4 max connections or the number of CPUs available.

bck01215:
https://github.com/jackc/pgx/blob/v5.7.1/pgxpool/pool.go\#L298-L324 shows an example on how to increase max connections

### DIFF
--- a/lessons/209/go-app/config.go
+++ b/lessons/209/go-app/config.go
@@ -63,6 +63,9 @@ type DbConfig struct {
 
 	// Database to store images.
 	Database string `yaml:"database"`
+
+	// Max connections to the database.
+	MaxConnections int `yaml:"maxConnections"`
 }
 
 // loadConfig loads app config from YAML file.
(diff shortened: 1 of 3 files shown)


antonputra:
@bck01215 Thanks, that's useful! What is the typical pool size you use in production?

antonputra (approved):

antonputra:
@bck01215 If I'm not mistaken, by default, Postgres allows 100 connections, so this would limit it to 5 replicas max?

bck01215:
> @bck01215 Thanks, that's useful! What is the typical pool size you use in production?

It varies on what the app is doing in the database and how many CPUs are available to it. I typically start with 20, but probably for testing it would make sense to give both apps the same number of connections.


It looks like DJango uses, [psycopg](https://docs.djangoproject.com/en/5.1/ref/databases/#connection-pool) which should default to [4 connections as well](https://www.psycopg.org/psycopg3/docs/api/pool.html) but that  wasn't observed, so Maybe DJango overwrites the defaults. 

If you updating the settings.py DB pool to be a dict you should be able to explicitly overwrite the defaults. You can keep increasing the number until some other part of the app begins to to throttle.
I think this would work
```
            'pool': {
                'max_size': 20
            }
```


Postgres does default to 100 max connections, so that should be a hard limit set if you want to grow above 20.

PR #287, opened 2024-09-25
ref: node app using stdlib

vorandrew:
Let's compare apples to apples. NodeJS stdlib implementation vs  Golang stdlib


antonputra:
@vorandrew sure, will do! just to clarify, it won't matter how many CPU cores i give to node, right? for example, if i give it 4 CPUs, it will still use 1 because it's single-threaded? so again, it wouldn't be fair to compare, i guess. would it make more sense to create 2 replicas of each app and give each 1 CPU instead so we still get 2 CPU per application?

agravelot:
Node can use multiple cores for specific workloads.

https://www.youtube.com/watch?v=zphcsoSJMvM

antonputra:
@agravelot yes i saw it in documentation

vorandrew:
Yes, node.js spawns four threads in addition to the main thread but none of them are used for network I/O

vorandrew:
> @vorandrew sure, will do! just to clarify, it won't matter how many CPU cores i give to node, right? for example, if i give it 4 CPUs, it will still use 1 because it's single-threaded? so again, it wouldn't be fair to compare, i guess. would it make more sense to create 2 replicas of each app and give each 1 CPU instead so we still get 2 CPU per application?

It's not CPU cores you are giving... it's CPU limits. Go concurrency is limited by number of physical cores on machine or GOMAXPROCS

Here is good read  https://www.ardanlabs.com/blog/2024/02/kubernetes-cpu-limits-go.html

antonputra (approved):
@vorandrew Thanks! I meant limit.

lessons/210/node-app/package.json
```diff
@@ -11,7 +11,6 @@
   "license": "ISC",
   "description": "",
   "dependencies": {
-    "express": "^5.0.0",
     "js-yaml": "^4.1.0",
```
polRk, line 14: remove to

PR #288, opened 2024-09-25
ref: GOMAXPROCS=1 

vorandrew:
Due to concurrency nature of Golang you should also set GOMAXPROCS=1 to make results comparable, otherwise Go uses threads = max cores to serve requests but node uses only 1 thread. Also take notice if you were running on instances with more cores Golang results would be even higher with same CPU limits


antonputra:
@vorandrew well, actually, i was thinking of giving each application 1 CPU and instead running 2 replicas. i left another question in https://github.com/antonputra/tutorials/pull/287

antonputra:
@vorandrew @agravelot Last question: I just want to make it as 'fair' as possible. I use m7a.xlarge EC2 instances and deploy each application on its own instance. I set the CPU limit to 2000m and 256Mi of memory. I understand that by setting GOMAXPROCS=1, I potentially reduce the performance of Go. Is that fair to say? What setup would you suggest to make it fair without explicitly limiting Go's performance?

vorandrew:
Golang uses concurrency vs single threaded NodeJS. On top of this,  internal goroutine scheduler makes it even more efficient rather then system threads scheduler

vorandrew:
IMO. You should set same cpu limit for both apps - it will make both apps even in terms of throttling. On top of this you should neutralise concurrency effect of golang by setting GOMAXPROCS=1

antonputra:
@vorandrew I understand that, but that's the whole point of Go, isn't it? I can make it 'fair' on one side, but then I'll get a lot of criticism from the Go community 😊.

agravelot:
This is a good reference for a small system, which is not very representative of high-demand systems.
For highly scalable systems, you can use multi nodejs replicas against 1-2 replicas for golang. As long as your total limits "match".

In any case, the benchmarking methodology must answer a specific question. Otherwise, there is an answer for a multitude of specific uses.

vorandrew:
IMO. Golang is beating NodeJS in performance because it's concurrency model is better. You can nerf this to make comparison as close as possible, but should we?

antonputra:
I don't think it's a big deal. I'll just run two tests and share the results, one with GOMAXPROCS=1 and one without. Let's see how Node's standard library compares to Go.

antonputra (approved):

lessons/210/deploy/go-app/2-deployment.yaml
```diff
@@ -37,6 +37,9 @@ spec:
       containers:
       - name: go-app
         image: aputra/go-app-210:v3
+        env:
+        - name: GOMAXPROCS
+          value: 1
```
DarVoid, line 42: - name: GOMEMLIMIT
  valueFrom:
    resourceFieldRef:
      resource: limits.memory
- name: GOMAXPROCS
  valueFrom:
    resourceFieldRef:
      resource: limits.cpu

DarVoid: https://blog.howardjohn.info/posts/gomaxprocs/

antonputra: thanks!

vorandrew:
https://go.dev/blog/container-aware-gomaxprocs

PR #291, opened 2024-09-26
Lesson 210: improve Node.js app performance by usage of Fastify and uWebSockets.js

pasha-vuiko:
improve performance by usage of Fastify and uWebSockets.js (alternative node.js http and ws server which is more performant than node:http)


antonputra:
@pasha-vuiko Thank you for the PR. Could you please move it to the fastify-app folder and keep the original code in place for future reference? I'll merge it afterward.

pasha-vuiko:
@antonputra done

antonputra (approved):

PR #296, opened 2024-10-03
perf: use static option in bun server

mtt-artis:
hello 👋 
In tutorial 212 - Node vs Bun, I'm not sure if using Bun-specific features would be considered cheating.
https://bun.sh/docs/api/http#static-routes

Cannot wait to see how deno2 will perform


antonputra (changes requested):
Thanks @mtt-artis for the PR. Those routes shouldn't be static. The health check should verify the status of the application, and /api/devices should return data from the database. But I get your point. If you'd like, you can create a bun-app-static with a single app.js file and those two routes, and I'll merge it. It would be nice to have for future reference.

mtt-artis:
I agree. Using the static option in bun-app won't provide an accurate representation of the performance of a real HTTP server in Bun.

antonputra (approved):

PR #299, opened 2024-10-05
Tell golang about memory limit and allow it to manage the 2 cores itself

Kavantix:
Set GOMEMLIMIT to ca 95% of actual limit to allow for a bit of overhead by the system

I used the requests to do so to make it easier to configure


antonputra:
@Kavantix  oh no, I deleted the 213 branch, and it automatically closed this PR? Can you please create another one from main?

PR #301, opened 2024-10-05
fix(golang): Use same context as request

hkupty:
This would make a timeout in the request cancel the lingering db transaction


antonputra (approved):
Thanks, @hkupty, I missed it

hkupty:
Glad to help man. Thanks for the amazing content!

PR #303, opened 2024-10-06
Tell golang about memory limit and allow it to manage the 2 cores itself

Kavantix:
Set GOMEMLIMIT to ca 95% of actual limit to allow for a bit of overhead by the system

Same as #299 but based on main

### DIFF
--- a/lessons/213/k8s/deploy/go-app/1-deployment.yaml
+++ b/lessons/213/k8s/deploy/go-app/1-deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   name: go-app
   namespace: default
 spec:
-  replicas: 2
+  replicas: 1
   strategy:
     type: Recreate
   selector:
@@ -30,13 +30,17 @@ spec:
           valueFrom:
             resourceFieldRef:
               resource: limits.cpu
+        - name: GOMEMLIMIT
+          valueFrom:
+            resourceFieldRef:
+              resource: requests.memory
         resources:
           requests:
-            memory: 128Mi
-            cpu: 500m
-          limits:
-            memory: 256Mi
+            memory: 480Mi
             cpu: 1000m
+          limits:
+            memory: 512Mi
+            cpu: 2000m
         readinessProbe:
           httpGet:
             path: /healthz


sshilko:
With this GOGC can be adjusted or turned off altogether also, but this will be cheating against any other go app defaults, so probably not a good idea for comparable benchmarks cc @antonputra 

GOMEMLIIMT will trigger GC when needed, so GOGC can be turned off now in practice

lessons/213/k8s/deploy/go-app/1-deployment.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: go-app
   namespace: default
 spec:
-  replicas: 2
+  replicas: 1
```
sshilko, line 8: thats not very fair to bun to have 1 dedicated pod vs 2 smaller pods in bun

https://github.com/antonputra/tutorials/blob/main/lessons/213/k8s/deploy/bun-app/1-deployment.yaml

lessons/213/k8s/deploy/go-app/1-deployment.yaml
```diff
@@ -30,13 +30,17 @@ spec:
           valueFrom:
```
sshilko, line 30: trade all memory for minimal amount gc cycles?
```
- name: GOGC
   value: off
```

reduce gc cycles knowing ahead we will eat up all memory anyway ?
```
- name: GOGC
   value: 200
```


```
GOGC controls the aggressiveness of the garbage collector. By default this value is assumed to be 100, which means garbage collection will not be triggered until the heap has grown by 100% since the previous collection. Effectively GOGC=100 (the default) means the garbage collector will run each time the live heap doubles.

Setting this value higher, say GOGC=200, will delay the start of a garbage collection cycle until the live heap has grown to 200% of the previous size. Setting the value lower, say GOGC=20 will cause the garbage collector to be triggered more often as less new data can be allocated on the heap before triggering a collection.
```

lessons/213/k8s/deploy/go-app/1-deployment.yaml
Kavantix: I disagree, requiring 2 pods for bun is a limitation, running 2 pods for bon is arguably unfair but reasonable.

sshilko: This is due to "hidden" knowledge that Bun is single core. Okay

antonputra: I'll try bun workers next time but I'll definitely keep the same number of pods - https://bun.sh/docs/api/workers

antonputra:
Thanks @Kavantix for the PR and @sshilko for the feedback. I'll keep it as a reference, and when I have time, I'll try different GC settings.

antonputra (approved):

PR #304, opened 2024-10-08
fix(prometheus-nginx-exporter): prevent exporter crash on incorrect log format

hexley21:
In case in access.log file there is a line that does not match the regex, the app won't panic

So before adding match to a result map, it will check if length does not equal to 0


antonputra (approved):
Thanks for the fix

PR #310, opened 2024-10-12
Use FastHTTP in the Golang implementation

xplshn:


antonputra:
Thanks for the PR, @xplshn. If you can place it in fasthttp-app and keep the original for reference, I'll merge it right away.

antonputra (changes requested):
Please keep original app for reference and create "fasthhp-app"

xplshn:
Sounds good!

xplshn:
> Please keep original app for reference and create "fasthhp-app"

Already addressed :)

PR #315, opened 2024-10-14
213-go: make sure to close io.ReadCloser on saveDevice()

borring:
I'm not sure if it makes any difference to the test, but `http.Request.Body` does have an `io.ReadCloser` interface, so we should be making sure to close it after use.


antonputra (approved):
Thanks! I don't think it affects the test, but it's still better to close it

PR #319, opened 2024-10-18
Remove slow query log from MySQL

melroy89:
Despite that [I did indeed have slow query log enabled](https://gitlab.melroy.org/-/snippets/92) (for performance insides). In your case during executing benchmarks, I believe it's fair to disable slow query log fully. 

Since you don't do that for PostgreSQL either.


antonputra (approved):
noted

PR #325, opened 2024-10-27
Improve Nginx config

melroy89:
Me again 😆 .You are (again) using default configurations. Which is not good. 

I try to come-up with again a better Nginx configuration. I don't use Apache myself, but you most likely want to fine-tune that setup as well.

`m6a.xlarge` might be not large enough with only 2 cores to really stress the system.

For now I set the Nginx thread pool to use only 2 threads. However, if your test environment will have 4 or 8 cores, please increase the `threads` in the thread pool as well as max queue and `worker_connections`!

Last but not least, you're not running the latest Nginx version. Try to run nginx v1.27.2

### DIFF
--- a/lessons/219/nginx/nginx.conf
+++ b/lessons/219/nginx/nginx.conf
@@ -1,14 +1,61 @@
 user nginx;
 worker_processes auto;
+worker_cpu_affinity auto;
+# Change the default thread pool settings
+thread_pool default threads=2 max_queue=16384;
 
 error_log /var/log/nginx/error.log notice;
 pid /var/run/nginx.pid;
 
 events {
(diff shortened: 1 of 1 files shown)


antonputra (approved):
I appreciate your help!

antonputra:
@melroy89 You created PR against 219 not the main branch :(

melroy89:
You can git merge into main? Or should I create another pr?

melroy89:
> @melroy89 You created PR against 219 not the main branch :(

ow that is because I searched for the code. And only found it initially here: https://github.com/antonputra/tutorials/tree/219/lessons/219

PR #326, opened 2024-10-27
Improve Nginx config in main branch

melroy89:
Same as PR: https://github.com/antonputra/tutorials/pull/325

Please **update your link** in your video description..

### DIFF
--- a/lessons/219/nginx/nginx.conf
+++ b/lessons/219/nginx/nginx.conf
@@ -1,14 +1,61 @@
 user nginx;
 worker_processes auto;
+worker_cpu_affinity auto;
+# Change the default thread pool settings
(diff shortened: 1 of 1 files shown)


antonputra:
@melroy89 i specifically use git tag to reference exact commit that i used in the test

melroy89:
> @melroy89 i specifically use git tag to reference exact commit that i used in the test https://github.com/antonputra/tutorials/tags

I see, but be aware that people (like me) might then also use this link as starting point of editing or improving. Hence GitHub will create a PR against that same branch.

antonputra:
@melroy89 i used to provide a link to the main branch, but when I get a few PRs, it’s confusing which code was actually used in the test. however, I understand your point.

PR #327, opened 2024-10-29
Apache Performance Improvements  

x01AS:
# Apache Performance Improvements  

Hi Anton,  

In this PR, I’ve made the following changes to improve Apache performance:  

## httpd-mpm.conf  
- **`ThreadsPerChild`** increased to 64 (more aggressive for 2 vCPUs)  
- **`ServerLimit`** set to 8 (2 vCPUs * 4)  
- **`MaxRequestWorkers`** set to 512 (64 * 8)  
- **`MaxConnectionsPerChild`** set to 10,000, as you suggested (helps prevent memory leaks while maintaining good performance)  
- **`StartServers`** increased to 4 to handle initial load more efficiently  
- **`MinSpareThreads`** and **`MaxSpareThreads`** adjusted proportionally to manage load variations  

## httpd.conf  
- Added content type to optimize gzip settings  

## my-website.conf  
- Added **`ProxyPassReverse`** to ensure these URLs align with the client’s view of the application, avoiding confusion or access issues

### DIFF
--- a/lessons/219/apache/httpd-mpm.conf
+++ b/lessons/219/apache/httpd-mpm.conf
@@ -2,11 +2,26 @@
     PidFile "logs/httpd.pid"
 </IfModule>
 
+## Since testing on instance with 2vCPU, 
+## ThreadsPerChild increased to 64 (more aggressive for 2 vCPUs)
+## ServerLimit set to 8 (2 vCPUs * 4)
+## MaxRequestWorkers set to 512 (64 * 8)
+## MaxConnectionsPerChild set to 10000 as you suggested (helps prevent memory leaks while still allowing good performance)
 <IfModule mpm_event_module>
-    StartServers             3
-    MinSpareThreads         75
-    MaxSpareThreads        250
-    ThreadsPerChild         25
-    MaxRequestWorkers      400
-    MaxConnectionsPerChild   0
+    StartServers             4
+    MinSpareThreads        100
+    MaxSpareThreads        300
+    ThreadsPerChild         64
+    ServerLimit             8
+    MaxRequestWorkers      512
+    MaxConnectionsPerChild  10000
 </IfModule>
+
+# <IfModule mpm_event_module>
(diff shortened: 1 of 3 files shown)


antonputra (approved):
Thank you! I'll try it out, and if it makes a significant improvement, I'll update the benchmark.

melroy89:
Thank you for this, I'm not using Apache2 for 10 years. So I could only help with Nginx.

lessons/219/apache/httpd-mpm.conf
```diff
@@ -2,11 +2,26 @@
...
+## MaxRequestWorkers set to 512 (64 * 8)
+## MaxConnectionsPerChild set to 10000 as you suggested (helps prevent memory leaks while still allowing good performance)
 <IfModule mpm_event_module>
-    StartServers             3
-    MinSpareThreads         75
-    MaxSpareThreads        250
-    ThreadsPerChild         25
-    MaxRequestWorkers      400
-    MaxConnectionsPerChild   0
+    StartServers             4
```
melroy89, line 11: Just curious if this has any benefit, you will now run 4 processes, while you only have 2 cores (threads). It's not like having 2 literal cores, it's just have two 'threads'. Even with multi-threading the docker container will only have 2 threads to run the process. 

If 4 servers make sense, maybe I should increase the Nginx threads to 4 as well? https://github.com/antonputra/tutorials/pull/326/files#diff-b5388329a6f40563cb48fa67327f1fbe1dd347eb7db44012fa0ebfcf9a0c30baR5

Not sure.

melroy89: 4 servers with a max of 8 could actually be detrimental for performance when only using 2 vCPUs.

@antonputra Before you run a 4 or 8 hour test. You might want to briefly check which for both Nginx & Apache2 what the "sweet spot" would be in terms of number of threads/servers. And whether overallocation will benefit performance on Docker or not.

PR #328, opened 2024-10-30
Lesson 215: [golang] use json-iterator instead of encoding/json

helginet:
Switch to json-iterator a high-performance 100% compatible drop-in replacement of "encoding/json".

Benchmark


antonputra (approved):
Thanks! I'll try it out

PR #330, opened 2024-11-01
Lession 220: Ruby without Rails implementation

simpl1g:
I'm wondering if it is possible to test Pure Ruby. On my local M1 I see 25x improvement on get devices endpoint and 10x for create device.


antonputra (approved):
Thank you very much! I really appreciate your help. Rails is very slow... I'll test it in a couple of days.

simpl1g:
@antonputra Thank you! 

Just a quick question, do you use `http_*` metrics for prometheus? Or only `myapp_request_duration_seconds` for DB latency?
 
I will try to improve Rails performance today/tomorrow as well

antonputra:
@simpl1g No, I don't use internal HTTP metrics because it's hard to compare them between different applications. I use an external client that exposes HTTP metrics, which is more accurate and application-agnostic.

lessons/220/ruby-app/config.ru
```diff
@@ -0,0 +1,26 @@
...
+
+require_relative 'src/app'
+
+RubyVM::YJIT.enable
+Iodine.threads = ENV.fetch('RAILS_MAX_THREADS', 2).to_i
+Iodine.workers = ENV.fetch('WORKERS_NUM', 2).to_i
+Oj.default_options = { mode: :compat }
+pg_conf = YAML.load(File.read('db/config.yml'), aliases: true)['production']
+PG_POOL = ConnectionPool.new(size: Iodine.threads * Iodine.workers, timeout: 5) do
+  PG.connect(pg_conf['host'], pg_conf['port'], nil, nil, pg_conf['database'], pg_conf['username'], pg_conf['password'])
```
ermolaev, line 19: @simpl1g good pr, we can also optimize the work with database through prepared queries via mini-sql gem

https://github.com/discourse/mini_sql?tab=readme-ov-file#prepared-statements

```ruby
SQL = <<~SQL
  INSERT INTO "devices" ("uuid", "mac", "firmware", "created_at", "updated_at")
  VALUES (?, ?, ?, ?, ?)
  RETURNING "id"
SQL

PG_POOL.with do |mini_sql|
  device.id = mini_sql.prepared.query_single(SQL, device.uuid, device.mac, device.firmware, device.created_at, device.updated_at).first
end
```
prepared queries give a big win when there are a lot of joins, here I think we can get a 10% boost speed

simpl1g: @ermolaev thanks! I also thought about it, but node app doesn't use prepared statements, so I decided not to use them as well

antonputra: ok thanks, i used in some apps but not in node

antonputra:
@simpl1g when i run it with 2 workers `WORKERS_NUM=2 RAILS_MAX_THREADS=2 bin/iodine -p 8080 config.ru` I get the following error
```
objc[21226]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
objc[21226]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```
with a single worker it works just fine `WORKERS_NUM=1 RAILS_MAX_THREADS=2 bin/iodine -p 8080 config.ru`
is this expected behavior?

simpl1g:
If you running on mac you should add `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES`

```
WORKERS_NUM=2 OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES bin/iodine -p 8080 config.ru
```

antonputra:
@simpl1g Thank you for clarifying this

PR #334, opened 2024-11-03
Enable http keepalive for nginx upstream

yegor-usoltsev:
Enabling HTTP keepalive for upstream connections in NGINX can significantly improve performance in a reverse proxy scenario by:

- **Reducing CPU load** on upstream servers, as fewer HTTP connections need to be established.
- **Improving request latency** by reusing connections, which also aids in handling high request volumes.
  
This change aligns NGINX's behavior with Caddy and Apache, both of which enable keepalive by default. For more details, see:

- [reverse_proxy (Caddyfile directive) — Caddy Documentation](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#keepalive)
- [Apache Module mod_proxy_http](https://httpd.apache.org/docs/2.4/mod/mod_proxy_http.html)

References:

- [NGINX blog: 10 Tips for 10x Application Performance](https://www.f5.com/company/blog/nginx/10-tips-for-10x-application-performance#web-server-tuning) (Tip 9 – Tune Your Web Server for Performance)
- [NGINX blog: HTTP Keepalive Connections and Web Performance](https://www.f5.com/company/blog/nginx/http-keepalives-and-web-performance)
- [NGINX docs: ngx_http_upstream_module](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive)

### DIFF
--- a/lessons/219/nginx/my-website.conf
+++ b/lessons/219/nginx/my-website.conf
@@ -29,6 +29,7 @@ server {
 upstream myapp {
     server myapp-nginx-0.antonputra.pvt:8080;
     server myapp-nginx-1.antonputra.pvt:8080;
+    keepalive 64;
 }
 
 server {
@@ -41,5 +42,7 @@ server {
 
     location / {
         proxy_pass http://myapp/;
+        proxy_http_version 1.1;
+        proxy_set_header Connection "";
     }
 }
--- a/lessons/221/nginx/my-website.conf
+++ b/lessons/221/nginx/my-website.conf
@@ -17,6 +17,7 @@ server {
 upstream myapp {
     server api-nginx-0.antonputra.pvt:8080;
     server api-nginx-1.antonputra.pvt:8080;
+    keepalive 64;
 }
 
 server {
@@ -29,6 +30,8 @@ server {
 
     location / {
         proxy_pass http://myapp/;
+        proxy_http_version 1.1;
+        proxy_set_header Connection "";
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     }
 }


antonputra (approved):
thank you @yegor-usoltsev, i appreciate your help. i wonder why nginx doesn’t use it by default, though

melroy89:
Thanks for the keep alive information, I wasn't aware of this information and that Nginx by default closes every connection with the upstream service.

@yegor-usoltsev Does this mean that a simple proxy pass like this, is not sufficient?:

```nginx
    location / {
        proxy_set_header Host $host;
        proxy_set_header X-Forwarded-For $remote_addr;
        proxy_pass http://localhost:3000/; # Just an example port
    }
```

Of course this needs to be improved by setting:

```
        proxy_http_version 1.1;
        proxy_set_header Connection "";
```

And I need to create a separate `upstream` block for all these kind of proxies in order to leverage `keepalive`? I found to so strange that Nginx doesn't allow you to set `keepalive` in the location block itself..

yegor-usoltsev:
@melroy89

> And I need to create a separate `upstream` block for all these kind of proxies in order to leverage `keepalive`?

Short answer - **Yes**. Better keep a separate `upstream` block in order to use `keepalive` some small positive number.

Long answer - I've created a [repository](https://github.com/yegor-usoltsev/nginx-upstream-keepalive) where I explain in detail with examples why this is necessary and how it works. If you find helpful - give the repository a star ⭐

> EDIT: btw `keepalive` is not the max number of connections. So you don't need to set this value to a very high number! Often 10 or 30 should be sufficient already.

You're absolutely right, keeping the limit to around ~10 connections should be sufficient. The primary reason I selected 64 is that @antonputra, during his benchmarking process, will have hundreds if not thousands of simultaneous connections, and I wanted to ensure that we witness a clear improvement in performance in upcoming NGINX benchmarks 😄

PR #335, opened 2024-11-03
220: Improve Rails perf 5-10x

simpl1g:


antonputra (approved):
Thanks for the PR! Can you sum up where 5-10x performance coming from?

simpl1g:
Yes,
- it uses lightweight Fibers instead of Threads to handle concurrency, so it allows more db operations per process
- Removed unused middlewares
- Removed prometheus http collector, it adds around 20% and not used in node, send a patch to the prom client with improvement https://github.com/prometheus/client_ruby/pull/316, I guess it can be improved further

Final results for create endpoint, it is not 10x, only 3x RPS and 8x better latency
```bash
before:
wrk -c 100 -t 2 -d 10 --latency -s ruby-app/wrk.lua http://localhost:8080/api/devices
Running 10s test @ http://localhost:8080/api/devices
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   163.18ms  205.32ms   1.09s    80.02%
    Req/Sec     0.90k   210.50     1.21k    77.00%
  Latency Distribution
     50%    7.90ms
     75%  314.85ms
     90%  474.04ms
     99%  809.52ms
  17954 requests in 10.01s, 11.95MB read
Requests/sec:   1793.34
Transfer/sec:      1.19MB

after:
wrk -c 100 -t 2 -d 10 --latency -s wrk.lua http://localhost:8080/api/devices
Running 10s test @ http://localhost:8080/api/devices
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    20.20ms   11.63ms 137.45ms   91.93%
    Req/Sec     2.66k   647.28     3.45k    86.00%
  Latency Distribution
     50%   17.30ms
     75%   20.61ms
     90%   26.89ms
     99%   76.76ms
  52996 requests in 10.04s, 16.88MB read
Requests/sec:   5277.25
Transfer/sec:      1.68MB
```

antonputra:
Thanks again, @simpl1g! I'll compare it with Django in a week or so; Node.js doesn't make sense.

lessons/220/rails-app/app/controllers/devices_controller.rb
```diff
@@ -1,56 +1,34 @@
...
-      Device.new(id: 2, uuid: "d2293412-36eb-46e7-9231-af7e9249fffe", mac: "E7-34-96-33-0C-4C", firmware: "1.0.3", created_at: "2024-01-28T15:20:51.137Z", updated_at: "2024-01-28T15:20:51.137Z"),
-      Device.new(id: 3, uuid: "eee58ca8-ca51-47a5-ab48-163fd0e44b77", mac: "68-93-9B-B5-33-B9", firmware: "4.3.1", created_at: "2024-8-28T15:18:21.137Z", updated_at: "2024-8-28T15:18:21.137Z")
+      Device.new(id: 1, uuid: "9add349c-c35c-4d32-ab0f-53da1ba40a2a", mac: "5F-33-CC-1F-43-82", firmware: "2.1.6", created_at: "2024-5-28T15:21:51.137Z", updated_at: "2024-5-28T15:21:51.137Z").attributes,
+      Device.new(id: 2, uuid: "d2293412-36eb-46e7-9231-af7e9249fffe", mac: "E7-34-96-33-0C-4C", firmware: "1.0.3", created_at: "2024-01-28T15:20:51.137Z", updated_at: "2024-01-28T15:20:51.137Z").attributes,
+      Device.new(id: 3, uuid: "eee58ca8-ca51-47a5-ab48-163fd0e44b77", mac: "68-93-9B-B5-33-B9", firmware: "4.3.1", created_at: "2024-8-28T15:18:21.137Z", updated_at: "2024-8-28T15:18:21.137Z").attributes
```
ermolaev, line 9: @antonputra @simpl1g what is the practical meaning of this benchmark?
for example, in node.js just creates hashes, we can do the same in ruby, and we don't need ActiveRecord objects, hash is much lighter and faster

```ruby
devices = [
  { id: 1, uuid: "9add349c-c35c-4d32-ab0f-53da1ba40a2a", mac: "5F-33-CC-1F-43-82", firmware: "2.1.6", created_at: "2024-5-28T15:21:51.137Z", updated_at: "2024-5-28T15:21:51.137Z"}, 
  { id: 2, uuid: "d2293412-36eb-46e7-9231-af7e9249fffe", mac: "E7-34-96-33-0C-4C", firmware: "1.0.3", created_at: "2024-01-28T15:20:51.137Z", updated_at: "2024-01-28T15:20:51.137Z" },
  { id: 3, uuid: "eee58ca8-ca51-47a5-ab48-163fd0e44b77", mac: "68-93-9B-B5-33-B9", firmware: "4.3.1", created_at: "2024-8-28T15:18:21.137Z", updated_at: "2024-8-28T15:18:21.137Z" },
]
```

I think it's more correct to load records from postgresql here,  it is close to real use

```ruby
build_response(Device.find(1,2,3), 200)
```

simpl1g: I guess it is just CPU bound benchmark, to test how well it handles serialisation. 2nd test is for IO.
But I agree that comparing objects with plain hash is not ideal in this case

antonputra: @ermolaev well i'm going to test Rails with Django, and pure Ruby implementation from this PR with Node.js - https://github.com/antonputra/tutorials/pull/330

lessons/220/rails-app/app/controllers/devices_controller.rb
```diff
@@ -1,56 +1,34 @@
...
-    end
-  end
+    device = Device.new(params.slice(:mac, :firmware))
+    device.uuid = SecureRandom.uuid
 
-  # PATCH/PUT /devices/1
-  def update
-    if @device.update(device_params)
-      render json: @device
+    if device.save
```
ermolaev, line 20: when rails saving, query wrapped in a transaction, so here we have 3 queries to database

```ruby
  TRANSACTION (1.2ms)  BEGIN
  Device Create (1.7ms)  INSERT INTO "devices" ("uuid", "mac", "firmware", "created_at", "updated_at") VALUES ($1, $2, $3, $4, $5) RETURNING "id"  [["uuid", nil], ["mac", "dfdf"], ["firmware", nil], ["created_at", "2024-11-06 00:53:09.435080"], ["updated_at", "2024-11-06 00:53:09.435080"]]
  TRANSACTION (0.6ms)  COMMIT
```

this code is sensitive latency to database

I do not know how other frameworks (django,laravel) works, but they can only insertion (1 query)

we can write `device.send :create_or_update` for skip transaction
it improves performance  `Requests/sec: 6866.0961 -> 9428.1055` on my mac m1

ermolaev: @simpl1g @antonputra any ideas about this?

simpl1g: @ermolaev yeah, it makes sense to disable transactions if Django also is not using them. It is better to check Django before I guess. But you can at least create PR

antonputra: Let's keep it in place for now.

PR #336, opened 2024-11-03
lesson 221: enhance caddy config

mohammed90:
The config doesn't make for fair comparison because Caddy is configured to do extra work

### DIFF
--- a/lessons/221/caddy/Caddyfile
+++ b/lessons/221/caddy/Caddyfile
@@ -11,11 +11,8 @@ www-caddy.antonputra.pvt {
 	# Enable the static file server.
 	file_server
 
-    # Enable access logs similar to nginx/apache
-    log {
-        format transform `{request>remote_ip} - {user_id} [{ts}] "{request>method} {request>uri} {request>proto}" {status} {size}` {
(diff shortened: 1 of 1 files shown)


antonputra (approved):
i appreciate your help! i'll test it in the near future, and if it makes a difference, i'll refresh the benchmark.

lessons/221/caddy/Caddyfile
```diff
@@ -1,37 +1,31 @@
-www-caddy.antonputra.pvt:443 {
+www-caddy.antonputra.pvt {
 	# Set this path to your site's directory.
 	root * /data/my-website
 
-    # Enable compression.
-    encode gzip
+	# Enable compression.
+	encode gzip 3
```
francislavoie, line 6: This isn't right, it needs to be:

```
	encode {
		gzip 3
	}
```

Only simple encoder names can be on the argument line, if options are to be given to the encoder then it needs to go in subdirectives.

antonputra: Thanks @francislavoie, can you please PR?

PR #337, opened 2024-11-03
Fix indent

melroy89:
Indent is only for readability. But advised nevertheless.


antonputra (approved):
sure, but i have auto-format on save, and vscode produced that

melroy89:
> sure, but i have auto-format on save, and vscode produced that

ow haha. I see.

PR #345, opened 2024-11-08
feat: add elixir with bandit

yordis:
image


yordis:
btw; what directory I need to check to see how you are provisioning the infra?

antonputra (approved):
@yordis ok, i'll do it in about a week 😊. Infra - https://github.com/antonputra/tutorials/tree/main/lessons/196

antonputra:
@yordis By the way, your Dockerfile is broken. The final artifact is not executable.
```bash
-rw-r--r-- 1 nobody root   72 Dec  7 08:58 server
```

This is one way to fix it. Add this to the final stage.

```bash
RUN chmod +x /app/bin/server
```

lessons/222/elixir-app/lib/app/telemetry.ex
```diff
@@ -0,0 +1,25 @@
...
+  def init(_arg) do
+    children = [{Peep, [name: :app, metrics: metrics()]}]
+    Supervisor.init(children, strategy: :one_for_one)
+  end
+
+  def metrics do
+    [
+      distribution([:bandit, :request, :stop, :duration],
+        tags: [],
+        unit: {:native, :millisecond},
```
antonputra, line 20: @yordis  Also, this value needs to be in seconds, not milliseconds. First of all, it's a Prometheus (open telemetry) standard, and second, to visualize metrics on the same graph, they need to use the same time unit."

antonputra:
No need to create a PR; I already updated it and will release it with a new video.

lessons/222/elixir-app/lib/app/telemetry.ex
yordis: btw, here is the list of telemetry https://hexdocs.pm/bandit/Bandit.Telemetry.html

antonputra: Thanks

snatvb:
the image is cheat :D

antonputra:
> the image is cheat :D

What do you mean @snatvb?

snatvb:
> > the image is cheat :D
> 
> What do you mean @snatvb?

just a joke about a picture in pr message :)

antonputra:
@snatvb i thought docker image 😊

lessons/222/elixir-app/rel/vm.args.eex
```diff
@@ -0,0 +1,10 @@
+## Customize flags given to the VM: https://www.erlang.org/doc/man/erl.html
+## -mode/-name/-sname/-setcookie are configured via env vars, do not set them here
+
+## Increase number of concurrent ports/sockets
+##+Q 65536
+
+## Tweak GC to run more often
+##-env ERL_FULLSWEEP_AFTER 10
+
++sbwt none
```
yordis, line 10: I wish I could document Why I added `+sbwt none` since this does not come by default as part of the `mix release` and have a major impact on k8s environment.

Leaving a note here @antonputra I wish the impact of misconfiguring this one was showed case, nowadays is one of the very first things I check simply because I have seen and made the mistake myself far too many times

abc3: > have a major impact on k8s environment

@yordis Could you add more details about the impact? Have you noticed an increase in latency?

yordis: Check https://adoptingerlang.org/docs/production/kubernetes/#container-resources

chasers: Also a good one: https://stressgrid.com/blog/beam_cpu_usage/

PR #346, opened 2024-11-09
Switch node:http to uWS C++

OreQr:
µWebSockets.js is fast Node.js web server written in C++

Ruby was benchmarked with third-patry web server also written in C

Also downgrade node version to v22.11.0 because uWS doesn't support node 23

NOTICE: I didin't benchmark it locally but it's supposed to work.


antonputra:
@OreQr "I appreciate this, and I'll definitely test it. However, would you mind keeping the Node app as it is and creating another folder under the same lesson? Maybe something like 'node-app-websockets' or similar. I'll merge it right away.

antonputra (changes requested):
@OreQr "I appreciate this, and I'll definitely test it. However, would you mind keeping the Node app as it is and creating another folder under the same lesson? Maybe something like 'node-app-websockets' or similar. I'll merge it right away.

antonputra (approved):
Thanks!

PR #347, opened 2024-11-10
fix: peep bucket calculator setup

yordis:
Hey @antonputra I just realized I made a minor mistake 😭


antonputra (approved):
No problem, when i get to it i'll verify it anyway

PR #352, opened 2024-11-13
Add pricing:GetProducts to the Karpenter

amzar96:
Hi,

This PR aims to add an additional policy to the `controller-trust-policy.json` file in the Karpenter lesson. 

After successfully installing the Helm release, an error message from the Karpenter pod indicates that it is not authorized to perform the `pricing:GetProducts` action. As a result, Karpenter will use the pricing information from 2022.


antonputra (approved):
Thank you!

PR #353, opened 2024-11-14
Enable experimental fast proxy engine

kanaevad:
Hi.
Traefik from v3.2 supports new experimental fast proxy engine. Claims to handle 50% more requests per second.
Would be interesting to benchmark.
https://traefik.io/blog/traefik-proxy-v3-2-a-munster-release/


antonputra (approved):
Thank you @kanaevad for this! I'll definitely take a look and test it!

PR #362, opened 2024-11-28
Replace the unused GIN index with a BTREE index

SvetlinZarev:
Replace the unused GIN index with a BTREE index when searching by price, because the `->` operator is not accelerated by the GIN index.

Each index in PG supports only specific "operator classes" (see https://www.postgresql.org/docs/current/gin.html#GIN-BUILTIN-OPCLASSES) . When working with JSONB, GIN has two modes:
* `jsonb_ops` (default), which supports the `?`, `?|`, `?&``@>`, `@?`,`@@`,, etc operators
* `jsonb_path_ops`, which is smaller and faster, but supports only `@>`, `@?` and `@@`

Notice that none of those support `->`, which means that the GIN index will not be used when filtering by `jsonb -> 'price'`. The gin is great only for "contains" type of operations.

In order to fix that, I replace the GIN, which indexes the whole JSON body, with a BTREE, which indexes only the property we are interested in. This should improve insterts/updates/deletes, as the index is smaller  and faster to create/update. Also it will speed-up searches, because now the query will use the index to filter by price.

### DIFF
--- a/lessons/227/README.md
+++ b/lessons/227/README.md
@@ -10,7 +10,7 @@ CREATE TABLE product (
     jdoc jsonb
 );
 
-CREATE INDEX product_inx ON product USING GIN (jdoc);
+CREATE INDEX idx__product__price ON product using BTREE(((jdoc -> 'price')::NUMERIC));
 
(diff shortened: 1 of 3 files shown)


antonputra (approved):
Thank you @SvetlinZarev, I realized only after I ran this test. Good catch!

PR #367, opened 2024-12-07
Fix the inefficient usage of gnet

panjf2000:
Hi, great project over here!

However, the code of the 2nd test for `gnet` doesn't follow the [best practices](https://gnet.host/docs/tutorial/best-practices/), which brings down the performance significantly. Let's fix it!

Along with some code improvements.


antonputra (approved):
@panjf2000 I appreciate your help! I'll rerun a test in a week.
gnet vs fasthttp vs hertz vs net/http

PR #370, opened 2024-12-11
Make benchmarks with Elixir fairer

josevalim:
* The code previously used `Jason.encode!` but `Jason.encode_to_iodata!` should be preferred over IO devices. This should increase performance and reduce memory usage. This is what frameworks such as a Phoenix would have used by default

* The code was not encoding `App.Device` structs in the first test, but regular maps/dictionaries, which meant it could not preallocate the key data (an optimization I am certain the Go runtime performs)

* Ensure the datetime is encoded once in second test. The Go test was also encoding once, but the Elixir one did it four times (twice when encoding to JSON, twice when writing to the database)

* Bypass Ecto (a relational mapper) and use a direct query when inserting records. That's exactly what the Go benchmark does, which should make a large impact

* Allow :pool_count to be configured. If you are going to have a large pool size, the pool itself may become the bottleneck. If you want to have 500 connections on a machine with 8 cores, I recommend splitting them up, for example pool_size=64 * pool_count=8

I also refactored the code in the routes.ex to reduce duplication.

### DIFF
--- a/lessons/230/elixir-app/lib/app/device.ex
+++ b/lessons/230/elixir-app/lib/app/device.ex
@@ -12,6 +12,6 @@ defmodule App.Device do
         [uuid, mac, firmware, created_at, updated_at]
       )
 
-    id
+    %{device | id: id}
   end
 end
--- a/lessons/230/elixir-app/lib/app/device.ex
+++ b/lessons/230/elixir-app/lib/app/device.ex
@@ -1,12 +1,17 @@
 defmodule App.Device do
-  use Ecto.Schema
-
   @derive {Jason.Encoder, only: [:id, :uuid, :mac, :firmware, :created_at, :updated_at]}
-  schema "devices" do
-    field(:uuid, :string)
-    field(:mac, :string)
-    field(:firmware, :string)
-    field(:created_at, :utc_datetime_usec)
-    field(:updated_at, :utc_datetime_usec)
+  defstruct [:id, :uuid, :mac, :firmware, :created_at, :updated_at]
+
+  def save(device) do
+    %{uuid: uuid, mac: mac, firmware: firmware, created_at: created_at, updated_at: updated_at} =
+      device
+
+    %{rows: [[id]]} =
+      App.Repo.query!(
+        "INSERT INTO \"devices\" (uuid, mac, firmware, created_at, updated_at) VALUES ($1, $2, $3, $4, $5) RETURNING id",
+        [uuid, mac, firmware, created_at, updated_at]
+      )
+
+    id
   end
 end
--- a/lessons/230/elixir-app/lib/app/router.ex
+++ b/lessons/230/elixir-app/lib/app/router.ex
@@ -10,22 +10,16 @@ defmodule App.Router do
       |> Peep.get_all_metrics()
       |> Peep.Prometheus.export()
 
-    conn
-    |> put_resp_content_type("text/plain")
-    |> send_resp(200, metrics)
-    |> halt()
+    text(conn, 200, metrics)
   end
 
   get "/healthz" do
-    conn
-    |> put_resp_content_type("text/plain")
-    |> send_resp(200, "OK")
-    |> halt()
+    text(conn, 200, "OK")
   end
 
   get "/api/devices" do
     devices = [
-      %{
+      %App.Device{
         id: 1,
         uuid: "9add349c-c35c-4d32-ab0f-53da1ba40a2a",
         mac: "5F-33-CC-1F-43-82",
@@ -33,7 +27,7 @@ defmodule App.Router do
         created_at: "2024-05-28T15:21:51.137Z",
         updated_at: "2024-05-28T15:21:51.137Z"
       },
-      %{
+      %App.Device{
         id: 2,
         uuid: "d2293412-36eb-46e7-9231-af7e9249fffe",
         mac: "E7-34-96-33-0C-4C",
@@ -41,7 +35,7 @@ defmodule App.Router do
         created_at: "2024-01-28T15:20:51.137Z",
         updated_at: "2024-01-28T15:20:51.137Z"
       },
-      %{
+      %App.Device{
         id: 3,
         uuid: "eee58ca8-ca51-47a5-ab48-163fd0e44b77",
         mac: "68-93-9B-B5-33-B9",
@@ -51,20 +45,17 @@ defmodule App.Router do
       }
     ]
 
-    conn
-    |> put_resp_content_type("application/json")
-    |> send_resp(200, Jason.encode!(devices))
-    |> halt()
+    json(conn, 200, devices)
   end
 
   post "/api/devices" do
     try do
       {:ok, body, conn} = Plug.Conn.read_body(conn)
       body = Jason.decode!(body)
-      now = DateTime.utc_now()
+      now = DateTime.utc_now() |> DateTime.to_string()
 
       device =
-        App.Repo.insert!(%App.Device{
+        App.Device.save(%App.Device{
           uuid: Ecto.UUID.generate(),
           mac: body["mac"],
           firmware: body["firmware"],
@@ -72,23 +63,28 @@ defmodule App.Router do
           updated_at: now
         })
 
-      conn
-      |> put_resp_content_type("application/json")
-      |> send_resp(201, Jason.encode!(device))
-      |> halt()
+      json(conn, 201, device)
     rescue
       e ->
-        conn
-        |> put_resp_content_type("application/json")
-        |> send_resp(400, Jason.encode!(%{message: Exception.message(e)}))
-        |> halt()
+        json(conn, 400, %{message: Exception.message(e)})
     end
   end
 
   match _ do
+    text(conn, 404, "Not found")
+  end
+
+  defp text(conn, status, body) do
     conn
     |> put_resp_content_type("text/plain")
-    |> send_resp(404, "Not found")
+    |> send_resp(status, body)
+    |> halt()
+  end
+
+  defp json(conn, status, payload) do
+    conn
+    |> put_resp_content_type("application/json")
+    |> send_resp(status, Jason.encode_to_iodata!(payload))
     |> halt()
   end
 end
--- a/lessons/230/elixir-app/priv/repo/migrations/20241108023931_create_devices.exs
+++ b/lessons/230/elixir-app/priv/repo/migrations/20241108023931_create_devices.exs
@@ -2,13 +2,12 @@ defmodule App.Repo.Migrations.CreateDevices do
   use Ecto.Migration
 
   def change do
-    create table(:devices, primary_key: false) do
-      add(:id, :bigserial, primary_key: true)
+    create table(:devices) do
       add(:uuid, :string)
       add(:mac, :string)
       add(:firmware, :string)
-      add(:created_at, :timestamptz, null: false)
-      add(:updated_at, :timestamptz, null: false)
+      add(:created_at, :string, null: false)
+      add(:updated_at, :string, null: false)
     end
   end
 end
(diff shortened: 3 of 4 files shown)


josevalim:
Please let me know if you run the benchmarks again. In particular, I'd love to know why Elixir fails in the first test, as that was certainly unexpected, and I could not reproduce it on my machine (on my machine, Elixir gets 81k req/s after this PR, Go gets 110k req/s).

lessons/230/elixir-app/lib/app/device.ex
```diff
@@ -1,12 +1,17 @@
...
-  use Ecto.Schema
-
   @derive {Jason.Encoder, only: [:id, :uuid, :mac, :firmware, :created_at, :updated_at]}
-  schema "devices" do
-    field(:uuid, :string)
-    field(:mac, :string)
-    field(:firmware, :string)
-    field(:created_at, :utc_datetime_usec)
-    field(:updated_at, :utc_datetime_usec)
+  defstruct [:id, :uuid, :mac, :firmware, :created_at, :updated_at]
```
josevalim, line 3: Use regular structs, as in the Go benchmark.

lessons/230/elixir-app/lib/app/router.ex
```diff
@@ -51,44 +45,46 @@ defmodule App.Router do
...
     |> put_resp_content_type("text/plain")
-    |> send_resp(404, "Not found")
+    |> send_resp(status, body)
+    |> halt()
+  end
+
+  defp json(conn, status, payload) do
+    conn
+    |> put_resp_content_type("application/json")
+    |> send_resp(status, Jason.encode_to_iodata!(payload))
```
josevalim, line 87: Refactor and use the more efficient `Jason.encode_to_iodata!`.

lessons/230/elixir-app/priv/repo/migrations/20241108023931_create_devices.exs
```diff
@@ -2,13 +2,12 @@ defmodule App.Repo.Migrations.CreateDevices do
...
-    create table(:devices, primary_key: false) do
-      add(:id, :bigserial, primary_key: true)
+    create table(:devices) do
       add(:uuid, :string)
       add(:mac, :string)
       add(:firmware, :string)
-      add(:created_at, :timestamptz, null: false)
-      add(:updated_at, :timestamptz, null: false)
+      add(:created_at, :string, null: false)
+      add(:updated_at, :string, null: false)
```
josevalim, line 10: Once again, mirror Go.

lessons/230/elixir-app/lib/app/router.ex
```diff
@@ -51,44 +45,46 @@ defmodule App.Router do
...
   match _ do
+    text(conn, 404, "Not found")
+  end
+
+  defp text(conn, status, body) do
     conn
     |> put_resp_content_type("text/plain")
-    |> send_resp(404, "Not found")
+    |> send_resp(status, body)
+    |> halt()
```
josevalim, line 81: This one is just a refactoring to make the code cleaner.

lessons/230/elixir-app/lib/app/router.ex
yordis: There are the types of optimizations I avoided since I cared more about the "common" situations. Primarily because they aren't the default ones 😭 

I am glad you are aware of the test; now we get to see the impact of this extra stuff.

💜 🚀

josevalim: What do you mean this isn't the default one? That's precisely how Phoenix would encode it and it is the recommended way. I don't think there is any value in doing something different than what the most common web framework would do.

antonputra:
@josevalim I’ll rerun the test within a week. I was told that Kubernetes is not the best environment for Elixir. Would you recommend comparing them on VMs? Would an m7a.large be enough, or do I need a bigger instance? Anyway, please provide any recommendations so that the test can be fair to both.

antonputra (approved):

lessons/230/elixir-app/lib/app/router.ex
ruslandoga: FWIW I've also seen quite often that `Jason.encode!` was used instead of `Jason.encode_to_iodata!` in "user-space" code. I think the misconception comes from the name, people see `encode` and don't look any further ...

- https://github.com/search?q=Jason.encode_to_iodata%21+language%3AElixir&type=code
- https://github.com/search?q=Jason.encode%21+language%3AElixir&type=code

And I'm glad `:json` went with `encode` returning iodata by default :)

yordis: @josevalim It is not that important, I was curious about baseline of what I seem based on my career, before adding every optimization in the book. The obvious one is to remove Ecto, as an example.

I am also curious to add cowboy to the mix 🤔 just out of curiosity

josevalim:
@antonputra honestly, continue using k8s. If someone is evaluating Go vs Elixir, and they want to deploy to k8s, telling them to use a VM would most likely be not be accepted, so I believe you should equally refuse it. :) The most important is to assign a whole machine to both Elixir and Go. Otherwise, if you have a machine with 8 cores and you break that into 8 pods, that would be an inferior solution for both. :)

Btw, since you are here, are you using the Prometheus stats for anything? I am only asking because they may be measuring things, but depending on what you are doing with it, it is all the same.

antonputra:
@josevalim Ok, I'll keep using Kubernetes. I use m7a.large with 2 CPUs and deploy each app on its own VM with limits equal to VM capacity. Yes, I'm using internal Prometheus metrics exposed by each app to measure database latency, and I use an external client with Prometheus metrics as well to measure the overall latency of the requests. Also, cAdvisor for CPU/Memory usage.

antonputra:
@josevalim Knowing that, would you advise on these values? https://github.com/antonputra/tutorials/blob/main/lessons/230/deploy/elixir-app/0-deployment.yaml#L30-L34

josevalim:
@antonputra excellent questions. The max connections should be the same in both for fairness. So my suggestion would be POOL_COUNT=4 and POOL_SIZE=125 if the goal is 500. I think the overall number does not matter as long as it is the same in both.

When it comes to the queue target and interval, Elixir employs a load shedding algorithm to drop connections if it knows the system is overloaded. The idea is that, instead of piling requests and overloading the database, we drop some of them if we cannot fulfill a target. The general principle is that it is best to serve a high % of people in reasonable time rather than a high % of people with high latency.

However, because your benchmark is precisely to measure how long it takes for things to start falling over, those numbers are probably too low. I would set 60000 for both queue target and queue interval, since I assume Go is not load shedding.

lessons/230/elixir-app/lib/app/router.ex
josevalim: @yordis the issue is that, by not making them as close as possible, wrong conclusions may be derived from the data. Ecto should have not been used for inserts from the beginning IMO. :) Regarding Cowboy, Bandit has been in my experience consistently faster.

@ruslandoga unfortunately Elixir's new JSON mirrors Jason. The trade-offs here are tricky, because if `JSON.encode!` returned iodata, people would be very confused and it may even be the incorrect result in some cases. I have added a clear warning/disclaimer to the docs though. :)

antonputra:
@josevalim Thank you for your help! I'll keep it open for a couple of days so I don't forget to make those configuration changes.

lessons/230/elixir-app/lib/app/router.ex
yordis: The intent was to have multiple runs, I am interested in seeing the differences load test, that is all

yordis:
@josevalim should we also add `+S ${SCHEDULERS}` to the VM args?

josevalim:
That should be automatically figured out. If you want to be sure, have it print System.schedulers_online when the app starts.

yordis:
> With OTP-23, released in 2020, the Erlang VM is "container aware" and will s the proper number of active schedulers automatically based on the containers allocated resources. Before OTP-23 the +S argument was needed to set the number of active schedulers equal the container's CPU limit.

Ahh yeah uuuuff I thought I would forgot, but probably a good idea to make sure it is working as intended

lessons/230/elixir-app/lib/app/router.ex
gregors: @yordis @josevalim - I'd also like to see Cowboy as well. The reason I mention this is because in my own testing when hitting both servers with load testing scripts, I saw much higher CPU usage from Bandit than Cowboy. Like much much higher at the limit. This was about 9 months ago, I might see if I can reproduce this and if so might ask in the forum for some help trying to discover the reason. ....assuming I can reproduce it.

yordis: This is a rabbit hole that we love 😄 we will do all possible combinations to find the fastest one!

untoldbyte:
So even at the rerun, the numbers do not increase significantly.  🤔

yordis:
@untoldbyte I still see a lot of CPU throttling, and now I wonder if scheduling and the CPU are the issues. To be honest, I wasn't even expecting the initial outcome numbers.

josevalim:
@antonputra thank you for doing another run! Quick question: how do you measure timeouts? Is the client timing out or the server? If they are server based, do you still have the server logs? In case you don't, could you keep them around if you happen to run it again?

Also, another question: how often does Prometheus pings the running applications in both cases?

Suggested next steps are:

1. Update to Elixir v1.18 and use the new JSON encoder
2. Upgrade Peep (we pushed some performance improvements for large number of buckets - which this benchmark uses - but we are waiting on a new release)
3. Upgrade Bandit (just in case)

Although I am not expecting large changes to be honest. After all, we are comparing a compiled language vs a functional language running on a VM, so a reasonable difference within the same order of magnitude is expected (well, unless you are the Java VM with its decades of JIT optimizations). :)

Happy holidays! 🎄

lessons/230/elixir-app/lib/app/router.ex
josevalim: Bandit is generally faster than Cowboy, so I am not expecting large changes here.

yordis:
@josevalim By now, I would be interested in running the benchmark in a VM outside K8s to isolate the problem and identify any misconfiguration that could be associated with K8s.

The CPU is being throttled and will cause timeouts (as your curiosity is hinting), especially if it is smack down to lower usage to go back to 100% again and smack down and go back ... I prefer to see a 100% CPU usage but steady usage.

Maybe I am wrong; I am here to learn and double-check my assumptions.

antonputra:
@josevalim Actually, for this benchmark, I upgraded the Docker image to v1.18 as well as Bandit to the latest version. - https://github.com/antonputra/tutorials/blob/main/lessons/233/elixir-app/Dockerfile#L14

Timeouts are configured on the client side using a Rust client that uses the `reqwest` library. 
```rust
client.get(url).header(ACCEPT_ENCODING, "gzip,deflate").timeout(Duration::from_millis(config.test.request_timeout_ms)).send()
```

Prometheus metrics for CPU and memory usage are scraped from cAdvisor, not the application. Only database latency is scraped every 5 seconds from the app.

I don't think there's a point in rerunning the same test again, even on a VM, but if there's a use case where Elixir can outperform Go, I would be open to demonstrating it. However, it has to be simple enough so that it can be replicated in Go. If you are interested, @yordis, take a look at the use cases and compare them with Go.

stevebrainng:
Any update on this? did you run the test based on this PR? I would love to see that, as I'm in point of choosing which language for our realtime application backend.

josevalim:
Honestly, I don't believe you should put much emphasis on benchmarks. I would try different frameworks/stacks, read the docs, interact with the communities, and see which ones you and your team will be more productive with. You can find huge corporations running on Go, Elixir, Ruby, PHP, etc, regardless of what online benchmarks tell you. [Discord realtime messaging is actually built on Elixir](https://elixir-lang.org/blog/2020/10/08/real-time-communication-at-scale-with-elixir-at-discord/) and they run at a massive scale. Whatsapp also runs on Erlang VM (the platform Elixir runs on) and they have more than a billion users.

I am not saying this to defend Elixir, you can pick a language faster than Elixir or a slower one. My point is that benchmarks should not play a big role on your decision process and I wouldn't let it block you from progressing, regardless of what your choice is.

PR #372, opened 2024-12-12
Use asyncpg in Python (231)

AloizioMacedo:
This should make Python be able to leverage concurrency, at least regarding postgres. I did not change anything related to memcached.

I am currently on vacation and about to leave at this moment, so I won't be able to test things. (E.g., I am not 100% sure if the URI with asyncpg is that one.) Please tell if you face any problems with the PR, I will likely be able to read later today.

---

Quick explanation on the problem of the previous version:

By marking a route as `async` as was done with `create_device`, it will be run in the main thread, blocking other things when running non-async functions. Since your db driver is not async, every db call will block any other ones that might be incoming. This greatly prevents concurrency. It might be possible that even just dropping the async determination of the route would increase performance, as that would run routes in a threadpool instead of the event loop in the main thread.

However, once we use an async db driver, not only the problem above goes away, but we should be leveraging concurrency greatly. Of course, Python will still have a (likely huge) inferior performance, but now in a scenario where the concurrency model makes more sense at least.

### DIFF
--- a/lessons/231/deploy/python-app/0-deployment.yaml
+++ b/lessons/231/deploy/python-app/0-deployment.yaml
@@ -27,7 +27,7 @@ spec:
         - name: MEMCACHED_HOST
           value: "memcache.antonputra.pvt"
         - name: POSTGRES_URI
-          value: "postgresql+psycopg://python_app:devops123@postgres.antonputra.pvt/mydb"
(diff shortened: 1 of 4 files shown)


antonputra (approved):
Thank you @AloizioMacedo! I'll wait a few days for more improvements and i'll rerun the test.

PR #375, opened 2024-12-13
Replace SQLModel with SQLAlchemy and Optimize Query Performance

techUdayMungalpara:
### Key changes:
- Replaced SQLModel with SQLAlchemy models, as SQLAlchemy provides the necessary functionality directly.
- Optimized the device creation endpoint by improving the query and using a direct SQLAlchemy insert with the `RETURNING` clause.

These changes improve performance.  review and let me know if any further adjustments are required.

PR #376, opened 2024-12-13
fix: python performance improvements

NewSouthMjos:
1 - get rid from ORM (SQLAlchemy) completely
2 - proper use of RETURNING statement, like on Goland app
3 - proper remove access log (using uvicorn directly now)

### DIFF
--- a/lessons/231/python-app/Dockerfile
+++ b/lessons/231/python-app/Dockerfile
@@ -21,7 +21,7 @@ COPY --from=build /usr/local/bin /usr/local/bin
 COPY . /app
 
 # Match the number of CPU cores, which is 2 on m7a.large, for the 1st test to reduce throttling.
-# CMD ["uvicorn", "run", "--workers", "2", "main:app", "--no-access-log"]
+# CMD ["uvicorn", "main:app", "--workers", "2", "--no-access-log"]
 
 # 4 for the second test to achieve full CPU utilization
-CMD ["uvicorn", "run", "--workers", "4", "main:app", "--no-access-log"]
+CMD ["uvicorn", "main:app", "--workers", "4", "--no-access-log"]
--- a/lessons/231/python-app/Dockerfile
+++ b/lessons/231/python-app/Dockerfile
@@ -21,7 +21,7 @@ COPY --from=build /usr/local/bin /usr/local/bin
 COPY . /app
 
 # Match the number of CPU cores, which is 2 on m7a.large, for the 1st test to reduce throttling.
-# CMD ["fastapi", "run", "--workers", "2", "main.py"]
+# CMD ["uvicorn", "run", "--workers", "2", "main:app", "--no-access-log"]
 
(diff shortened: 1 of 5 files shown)


antonputra (changes requested):
Thanks @NewSouthMjos, would you mind to resolve conflicts. Someone before you just made a similar updates. https://github.com/antonputra/tutorials/pull/375

antonputra:
@NewSouthMjos 🙏

lessons/231/python-app/Dockerfile
```diff
@@ -21,7 +21,7 @@ COPY --from=build /usr/local/bin /usr/local/bin
 COPY . /app
 
 # Match the number of CPU cores, which is 2 on m7a.large, for the 1st test to reduce throttling.
-# CMD ["fastapi", "run", "--workers", "2", "main.py"]
+# CMD ["uvicorn", "main:app", "--workers", "2", "--no-access-log"]
 
 # 4 for the second test to achieve full CPU utilization
-CMD ["fastapi", "run", "--workers", "4", "main.py"]
+CMD ["uvicorn", "main:app", "--workers", "4", "--no-access-log"]
```
tjwald, line 27: you should use gunicorn

tjwald: uvicorn is for development

pkucmus:
Should we also not compare a std lib from GO with a Python framework? Depending on what we are after we should either take Starlette or even make a simple ASGI application. What do you think?

Also what will happen with this PR since the video is out now?

lessons/231/python-app/Dockerfile
pkucmus: Maybe to state a "why". In some limited tests I did in the past, Gunicorn took ~75% of the RAM Uvicorn took and produced ~158% requests per second compared to the same app ran on Uvicorn.

antonputra:
@pkucmus I'll wait for a few more days to collect all the feedback and release a round 2 version with updated Python code.

pkucmus:
> Comparing Go stdlib with Python frameworks is actually valuable, especially for educational purposes! The contrast between compiled vs interpreted languages helps demonstrate important concepts around performance and concurrency models. It's a great way for developers, especially newcomers, to understand performance characteristics and tradeoffs between different tech stacks.

This is a nice way to put it and I would not argue against the educational value, then again, it depends on the purpose/goal of the benchmarks. To show that something is faster than the other thing? I don't have the answer myself and maybe I'm locked into a specific way of thinking but I seem to be stuck on failing to find how comparing the two at hand provides a real life benefit i.e. helping to make a business decision about tech (aside from what I tried to describe here: https://github.com/antonputra/tutorials/pull/382#issuecomment-2543927483)

antonputra:
Thank you @NewSouthMjos for the PR. Since most of it has already been implemented and merged in other PRs, I'll close it for now.

PR #378, opened 2024-12-14
Use Python's `perf_counter` for benchmarking purposes.

RealistikDash:
This PR replaces all uses of `time.time()` with `time.perf_counter()`. This is because the `time.time()` function uses the system clock, meaning it is susceptible to things such as leap seconds and clock re-syncs, which may impact the results. `perf_counter` uses the same unit (seconds).

Realistically, this is unlikely to happen during testing and this change does not affect performance (tested myself), but it makes the results more reproducible.


antonputra (approved):
Thanks!

PR #379, opened 2024-12-14
remove async when not necessary

tjwald:
This fixes the number of requests processed by each worker of gunicorn. 
https://stackoverflow.com/a/71517830/4549616

### DIFF
--- a/lessons/231/python-app/main.py
+++ b/lessons/231/python-app/main.py
@@ -39,12 +39,12 @@ uvicorn_logger.addFilter(LogFilter())
 
 
 @app.get("/healthz", response_class=PlainTextResponse)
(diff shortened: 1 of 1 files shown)


antonputra:
@tjwald thanks for the PR, i would definitely need to test this

antonputra:
@techUdayMungalpara Okay, I'll trust you guys

antonputra (approved):

rcaillon-Iliad:
It looks like a regression to me. The [doc](https://fastapi.tiangolo.com/async/#in-a-hurry) states the following:

> If your application (somehow) doesn't have to communicate with anything else and wait for it to respond, use async def.

PR #381, opened 2024-12-15
Change list to tuple, add limiting for max threads

antonpetrov145:
Hello Anton, in this PR:
- Change response to tuple instead of list:
  - Doing this allows for faster read of the object as tuples are faster than lists
- Add limiting for max threads that can be used by the app;
- Add usage of uvicorn with `--no-access-log` option;
- Remove uvicorn backlog number of connections;

You can adjust this
```
limitter.total_tokens = 1000
```
to a lower or higher number depending on the machine you use.

### DIFF
--- a/lessons/231/python-app/Dockerfile
+++ b/lessons/231/python-app/Dockerfile
@@ -24,4 +24,4 @@ COPY . /app
 # CMD ["fastapi", "run", "--workers", "2", "main.py"]
 
 # 4 for the second test to achieve full CPU utilization
-CMD ["fastapi", "run", "--workers", "4", "main.py"]
+CMD ["uvicorn", "--workers", "4", "--no-access-log", "main:app"]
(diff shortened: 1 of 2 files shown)


antonpetrov145:
You are absolutely correct. I am just not quite good with explanations. My commits are geared towards the first test - the get endpoint.

The idea here is not to have it lowered - it is the opposite. By default we have only 40 threads to operate with. Also in the code currently the first endpoint is defined as `def get_devices` which tells fastapi to run it in a separate thread pool.

So if we have 100 requests you can see that some requests will wait for a thread to open so it can be consumed.

antonputra (approved):

antonputra:
Thanks @antonpetrov145 for the PR, we have a conflict. So what should we do here? @techUdayMungalpara

antonpetrov145:
ok nice, i guess this can be closed then

PR #382, opened 2024-12-15
[231] drop FastAPI in favor of Starlette

pkucmus:
I'd like to drop FastAPI in favor of Starlette which is smaller and should be a closer comparison to GO's std lib. 

I would even make the argument that we should write an ASGI app here - an `async def app(scope, receive, send)` function.

Python will still loose to GO but it would be interesting to see that in a more apples to apples comparison.

What do y'all think about a pure and frameworkless test on the Python side?

I new here I did not know what are the rules to follow. This "lesson" already has multiple SQL definitions (in a yml and in schema.sql). I've adjusted the code to the one without TZ in the timestamp fields. I'm also not sure if I'm doing the right thing with the cache_client.

I also switched to Gunicorn as my very basic tests with Locust showed me one time that Gunicorn took ~75% of the RAM Uvicorn took and produced ~158% requests per second compared to the same app ran on Uvicorn.


pkucmus:
@antonputra I'm not sure how versed you are in the Python "ecosystem" so I apologize if I'm wasting your time with the following:

To provide a bit of context, in Python we have the ASGI protocol which basically is "THE" method to transform HTTP into Python. An ASGI application is just one function that takes 3 specific arguments. Starlette is one such application and FastAPI is built on top of it. 
So basically Starlette implements that function somewhere, builds a request representation, does the routing, etc.

Then we have Gunicorn and Uvicorn acting as the ASGI servers (Gunicorn is old and only works with ASGI thanks to Uvicorn workers).

Now my question is if we should measure a pure ASGI implementation vs GO's std lib. Or are we more after real-life examples? If the latter then Starlette and even FastAPI are more accurate if GO's std lib is something people actually build GO web apps with. I hope I'm making sense.

antonputra:
Thank you @pkucmus for the explanation, it is very helpful. I understand that FastAPI is built on top of Starlette, but most people asked me to compare FastAPI since it's very popular. So first, I want to rerun the test with an updated FastAPI application. Second, I'm thinking about running a test that would include only 4 Python frameworks. Maybe FastAPI vs. Starlette vs. ASGI app vs. Flask; any feedback on which frameworks to choose is welcome. Since I'm using external dependencies, I can't really run more than 4 frameworks at the same time.

So, if you don't mind, can you please create another 'starlette' folder and place your implementation there so that I can compare them in the future? Also, if you still have time, can you please create an ASGI app as well? I don't think it's going to take a lot of time. If you decide to help with the ASGI, I promise it won't take longer than a week to run both tests and publish results.

pkucmus:
Alright. Since FastAPI should not be replaced, let's close this one and move the conversation to https://github.com/antonputra/tutorials/pull/386 which is about adding Starlette (instead of replacing FastAPI with it)

PR #385, opened 2024-12-15
Upgrade Production Server to Gunicorn and Uvicorn

techUdayMungalpara:
Upgrade Production Server to Gunicorn and Uvicorn

### DIFF
--- a/lessons/231/python-app/starlette_app.py
+++ b/lessons/231/python-app/starlette_app.py
@@ -1,50 +1,42 @@
 import datetime
 import logging
+import os
 import time
 import uuid
 
 import aiomcache
 import orjson
 from asyncpg import PostgresError
-from fastapi import FastAPI, HTTPException
-from fastapi.responses import ORJSONResponse, PlainTextResponse
+from starlette.applications import Starlette
+from starlette.requests import Request
+from starlette.responses import PlainTextResponse, Response
+from starlette.exceptions import HTTPException
+from starlette.routing import Route, Mount
 from prometheus_client import make_asgi_app
 from pydantic import BaseModel
 
-from db import MemcachedDep, PostgresDep, lifespan
+from db import db, lifespan, MemcachedClient
 from metrics import H
 
-app = FastAPI(lifespan=lifespan)
 
-
-metrics_app = make_asgi_app()
-app.mount("/metrics", metrics_app)
-
-# Disable access logs to match Go implementation
-block_endpoints = ["/api/devices", "/metrics/", "/metrics"]
-
-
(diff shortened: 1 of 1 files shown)


lessons/231/python-app/db.py
```diff
@@ -1,17 +1,23 @@
...
 import asyncpg
-from fastapi import Depends, FastAPI
+from fastapi import Depends, FastAPI, HTTPException
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 
-POSTGRES_URI = os.environ["POSTGRES_URI"]
+POSTGRES_URI = POSTGRES_URI = os.environ["POSTGRES_URI"]
```
pkucmus, line 14: Typo?

lessons/231/python-app/main.py
```diff
@@ -102,13 +100,25 @@ async def create_device(device: DeviceRequest, conn: PostgresDep):
...
+            "created_at": now,  #
+            "updated_at": now,
+        }
 
         # Measure cache operation
         start_time = time.perf_counter()
 
         await cache_client.set(
-            device_uuid.bytes,
+            str(device_uuid).encode("utf-8"),
```
pkucmus: `device_uuid.hex.encode()` should be lighter

lessons/231/python-app/main.py
```diff
@@ -86,43 +86,42 @@ async def create_device(
         now = datetime.datetime.now(datetime.timezone.utc)
         device_uuid = uuid.uuid4()
 
-        insert_query = """
-            INSERT INTO python_device (uuid, mac, firmware, created_at, updated_at)
-            VALUES ($1, $2, $3, $4, $5)
-            RETURNING id ;
+        async with conn.transaction():
```
pkucmus, line 89: I don't think this is ok, for one: the GO implementation does not seem to operate in a transaction (I might be wrong, I weak at GO). Secondly this runs into the connection pool problem, you only need a connection for line 97, here you will return the connection on the exit (if that's what the transaction is for here) or on the exit from the Dependency. Also there's no commit.

lessons/231/python-app/main.py
```diff
@@ -115,6 +115,8 @@ async def create_device(
                 "updated_at": now,
             }
 
+            # Transaction ends here - automatic commit if successful, rollback if error
```
pkucmus, line 118: This is fine but is it returning the connection to the connection pool. I'm nagging about this as there's a possibility to limit the whole application's concurrency to just the `POSTGRES_POOL_SIZE`.
Still there doesn't seem to be a transaction on the GO's side

PR #384, opened 2024-12-15
Add Memcached connection pooling and transaction handling 

techUdayMungalpara:
Add Memcached connection pooling

• Use MemcachedDep in /api/devices endpoint
• Handle connection & transactions in pool
• Added connection pooling
• Enable monitoring stats

Test completed:
• Memcached ops & DI ✓
• PostgreSQL integration ✓

PR #386, opened 2024-12-15
feat: Starlette aside from FastAPI

pkucmus:
@antonputra to do what you asked for in https://github.com/antonputra/tutorials/pull/382#issuecomment-2543983150 I'd like to propose the following, here you have a starlette_app.py module with a Starlette counterpart of the FastAPI-based main.py. Since main.py is still being worked on, I'd like to pause here and wait with the other approaches. Once ready we can create more Python approaches based on how main.py will look like. Does that sound like a plan?

I took the liberty to touch up what was proposed by @techUdayMungalpara and just merged in https://github.com/antonputra/tutorials/pull/384. 

One additional thing I would like to point out (again) is how I release the connection back to the pool https://github.com/pkucmus/antonputra-tutorials/blob/7bdf33375279504e1272dfeab0526067a053c8ca/lessons/231/python-app/starlette_app.py#L89-L92, can we incorporate that into the FastAPI one as well?

### DIFF
--- a/lessons/231/python-app/starlette_app.py
+++ b/lessons/231/python-app/starlette_app.py
@@ -1,50 +1,42 @@
 import datetime
 import logging
+import os
 import time
 import uuid
 
 import aiomcache
 import orjson
 from asyncpg import PostgresError
-from fastapi import FastAPI, HTTPException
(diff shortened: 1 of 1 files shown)


antonputra (approved):
Makes sense, thank you Paweł!

PR #387, opened 2024-12-16
Update Gunicorn for Production

techUdayMungalpara:
Add Gunicorn to requirements.txt
Update Dockerfile to use Gunicorn
Convert lists to tuples for better performance (per @antonpetrov145  )

### DIFF
--- a/lessons/231/python-app/Dockerfile
+++ b/lessons/231/python-app/Dockerfile
@@ -24,4 +24,11 @@ COPY . /app
 # CMD ["fastapi", "run", "--workers", "2", "main.py"]
 
 # 4 for the second test to achieve full CPU utilization
-CMD ["fastapi", "run", "--workers", "4", "main.py"]
+CMD ["gunicorn", 
+     "-w", "4", 
+     "-k", "uvicorn.workers.UvicornWorker", 
+     "--timeout", "60", 
+     "--graceful-timeout", "60", 
+     "--log-level", "error", 
+     "main:app", 
+     "--bind", "0.0.0.0:8000"]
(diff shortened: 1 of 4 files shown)


antonputra:
@techUdayMungalpara I thought Unicorn is the recommended server for FastAPI, which works with ASGI. Am I missing something? cc @antonputra @pkucmus

antonputra (approved):

antonpetrov145:
@antonputra yes you are correct, even `fastapi run` runs uvicorn under the hood.

https://www.uvicorn.org/#running-with-gunicorn 
Reasoning behind using `gunicorn` is because it is more mature and offers more control over the main and worker processes despite it being WSGI server it is better to use it in production instead of pure uvicorn.

For instance in `uvicorn` you can't tell it to recreate a workter process after a number of requests, or how many max requests each worker to consume before being recreated.

Tradunsky:
Great PR! ❤️ 

Awesome videos @antonpetrov145 👏🏻 


Just want to add that web server impacts performance of a web framework, not just web framework gaining performance alone. 

With that in mind, based on my personal experience, modern web servers actually beat gunicorn quite noticeably. 

One that I tried and resulted in better latency and throughput is https://github.com/emmett-framework/granian/tree/master/benchmarks

Generally speaking, to get a good idea on the web servers progress you can look at the following benchmark:
https://www.techempower.com/benchmarks/#hw=ph&test=fortune&section=data-r22&p=zik0zj-zik0zj-zik0zj-zik0zj-qmx0qn-sf

That being said, not all web servers are production ready, some might be developed purely for the sake of blowing benchmarks. 

Hope it helps 🤗 

Happy to open up a PR if required.

PR #388, opened 2024-12-16
fix: remove unnecessary logging filter

NewSouthMjos:
Since the code now uses `--log-level error`, we no longer need an additional logger filter. Access-level logs are not captured anyway, and using a custom filter can slightly slow down the code.
Sorry for late response for previous PR.


seomago (approved):

antonputra (approved):
Thanks for the explanation @NewSouthMjos!

PR #389, opened 2024-12-17
Improve Python implementation in general.

RealistikDash:
This PR covers a lot of ground. Optimised using my own experience. 

## Summary of changes
- Re-factored the connection classes to not require an `self.pool is not None` check for every connection.
- Add a `__slots__` property to the connections. This simplifies attribute access for the interpreter (at the cost of not being able to add any attributes not mentioned in there).
- Reuse the Prometheus lib's labels per request. These actually turned out to be really expensive and do a lot (locking etc).
- Use the logging lib's formatting and exception logging. This means that both are done as needed rather than immediately before the function call.
- General low-hanging fruit, such as redundant `str` calls.

### DIFF
--- a/lessons/231/python-app/db.py
+++ b/lessons/231/python-app/db.py
@@ -131,7 +131,7 @@ async def lifespan(app: FastAPI):
         # Shutdown: close all connections
         logger.info(" Shutting down database connection...")
         await db.close()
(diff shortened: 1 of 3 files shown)


antonputra (approved):
Thanks @RealistikDash for the help!

PR #391, opened 2024-12-19
Improve fastapi perfs

rcaillon-Iliad:
- return Response objects to avoid extra overhead (https://fastapi.tiangolo.com/advanced/custom-response/#use-orjsonresponse)

> For large responses, returning a Response directly is much faster than returning a dictionary.

> This is because by default, FastAPI will inspect every item inside and make sure it is serializable as JSON, using the same [JSON Compatible Encoder](https://fastapi.tiangolo.com/tutorial/encoder/) explained in the tutorial. This is what allows you to return arbitrary objects, for example database models.

> But if you are certain that the content that you are returning is serializable with JSON, you can pass it directly to the response class and avoid the extra overhead that FastAPI would have by passing your return content through the jsonable_encoder before passing it to the response class.

- declare routes as async (https://fastapi.tiangolo.com/async/#in-a-hurry)

> If your application (somehow) doesn't have to communicate with anything else and wait for it to respond, use async def.

- call `orjson.dumps` one time only

### DIFF
--- a/lessons/232/python-app/main.py
+++ b/lessons/232/python-app/main.py
@@ -25,12 +25,12 @@ logger.setLevel(logging.DEBUG)
 
 
 @app.get("/healthz", response_class=PlainTextResponse)
-def health():
+async def health():
     return PlainTextResponse("OK")
 
 
 @app.get("/api/devices", response_class=ORJSONResponse)
-def get_devices():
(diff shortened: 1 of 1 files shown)


antonputra (approved):
Thanks

PR #392, opened 2024-12-20
feat: litestar application

NewSouthMjos:
Implemented Litestar framework application.
Used all the current implementation of database and memcache.

Litestar can be run with gunicorn too, for example:
```gunicorn -w 1 -k uvicorn.workers.UvicornWorker litestar_app:app --bind 0.0.0.0:8000```

So, Dockerfile needs to be adjusted


antonputra (changes requested):
Thanks @NewSouthMjos! Can you please put it in its own folder under the 232 lesson, maybe called 'litestar-app'? It's just easier for me not to get mixed up. Also, if you don't mind, could you add the updated Dockerfile? There's less chance that I'll make a mistake.

antonputra:
Or at least add `Dockerfile.litestar` in the root

NewSouthMjos:
Moved to lesson 232.
Note: litestar use mgspec library. Mgspec currently does not support python 3.13: https://github.com/litestar-org/litestar/issues/3265

Buy now, litestar-app using python 3.12

antonputra (approved):
Thank you! I'll test it in near future!

PR #394, opened 2024-12-22
Update vm.args.eex

Sur-en:
Enable SMP automatically based on availability.
This is first thing which guides recommend to do.
Also keep in mind that BEAM works better with concurrent model, which is mean that some sort of CPU suites better than others, for example last benches show that AMD CPUs works much more effectively with multicore utilisation and Intels better for single core performance. Moreover, BEAM usually runs without Kubernetes or Docker if performance really is must have. Overall Kubernetes written on Go and much more suites to Go than to Elixir, also add some latency and mostly solve same issues which solved in BEAM decades ago.


antonputra (approved):
Thanks! I was thinking of running the second test on a VM.

PR #399, opened 2025-01-07
feat: replace go-redis with rueidis

nesty92:
It will be interesting to see the results of Redis with Rueidis

https://github.com/rueian/rueidis-benchmark

client_test_set_10


nesty92:
Hi @antonputra, probably this is a good v2 Redis VS Memcache with go

antonputra (approved):
thanks! i'll gie it a try

PR #403, opened 2025-01-09
Replace fastapi with robyn for raw performance

antonpetrov145:
I know it is not what the test is about but when in comes to raw performance I haven't seen better framework than Robyn. It is a relatively new framework that has Rust core, so this will benefit the tests. Also it is marvelous and very similar to fastapi in terms of syntax.

IMO main "issue" with FastAPI and other python frameworks is that we are sacrificing raw performance for ease of use and rapid development. 

So in this PR:
- lowered the python version because 3.13 is not supported yet in robyn
- adapted the endpoints to use robyn
- adapted responses to robyn syntax 

If you are good with this change I will continue to push here so all database and memcached is working as expected.

### DIFF
--- a/lessons/236/fastapi-app/Dockerfile
+++ b/lessons/236/fastapi-app/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.13.1-slim-bookworm AS build
+FROM python:3.12.1-slim-bookworm AS build
 
 ENV PYTHONDONTWRITEBYTECODE=1
 ENV PYTHONUNBUFFERED=1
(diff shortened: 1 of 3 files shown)


antonputra (changes requested):
@antonpetrov145, thanks for the PR. I'd like to keep FastAPI as is. If you don't mind, could you please create another folder for the Robyn app or just another Python file? I already have a Starlette implementation, and I'll test FastAPI vs. Robyn vs. Starlette.

Starlette - https://github.com/antonputra/tutorials/blob/699af5dce52aa5e76a22c9a75978cbb3da78e564/lessons/232/python-app/starlette_app.py#L10

antonpetrov145:
@antonputra done, apps and files are separated

antonputra (approved):
Thanks!

PR #406, opened 2025-01-10
Use falcon host

mohammed-io:
Looks like the current version was running Iodine and Rage by default (which is not Rails + Falcon)

Screenshot 2025-01-10 at 7 16 27 PM

I have upgraded to the latest Rails version and Ruby 3.4.1 and used the recommended way of hosting it according to https://socketry.github.io/falcon/guides/deployment/index.html#configuration

Let me know if you need anything else or have questions!

Hope this helps!


antonputra (approved):
Thanks!

PR #416, opened 2025-01-28
Update 15-aws-lbc.tf

anujthedevopsguy:
when I tried this example, alb controller was not able to find the associated VPC id and throwing error. This is the fix of that issue


antonputra (changes requested):
thanks!

antonputra (approved):

PR #422, opened 2025-02-06
Fix the performance issues of teh rust application

SvetlinZarev:
# Proposed changes

1. Not speed related, but annoys me: there is no point is using `panic="abort"`, because it's not suitable for a productive webapp. I also doubt that it would give any significant speedup

2. Implement support for keep-alive. The bottleneck, at least on my machine was opening TCP connections. In the old implementation, the connectionw as closed after each request, but http clients such as apache http client, HttpUrlConnection, others, and of course the browsers, can cache the connection to the server in order to reuse it. As it turns out, opening a new TCP connections, even without TLS is quite slow. I'm also pretty sure that Golang does this out of the box, as it has a real HTTP server  unlike the toy server in the rust app.

3. Use buffered writers for writing to the stream, Some screenshots from my machine:

1. The original app
Screenshot 2025-02-06 at 20 36 56

2. After implementing keep-alive support
Screenshot 2025-02-06 at 20 38 18

3. After using buffered writer
Screenshot 2025-02-06 at 20 56 37

As you can see from the CPU graph, there is still a lot of room for improvement, but this PR at least makes the rust app competitive

### DIFF
--- a/lessons/242/rust-app/Cargo.toml
+++ b/lessons/242/rust-app/Cargo.toml
@@ -10,4 +10,3 @@ serde_json = "1.0.138"
 [profile.release]
 lto = true
 codegen-units = 1
-panic = "abort"
--- a/lessons/242/rust-app/src/routes.rs
+++ b/lessons/242/rust-app/src/routes.rs
@@ -1,5 +1,6 @@
 use crate::device::Device;
 use serde_json;
+use std::io::BufWriter;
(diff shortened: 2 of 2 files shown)


antonputra (approved):
Thank you!

melroy89:
@SvetlinZarev  where? I think you already found where. xD

PR #423, opened 2025-02-06
(bugfix) Terminate the connection even in case of keep-alive in order to be able to process more concurrent requests than available threads

SvetlinZarev:
Fix: in case of more concurrent connections that threads in the thread pool, the waiting connections will never be processed. So we terminate the currently running ines in order to rotate them.

The larger value we give to the config parameter, the better the throughput will be, but also the latency will be higher as well.

This is not suitable for productive usage, but is a requirement for this toy server in order to not block the additional connections forever during the tests.

For achieving the best performance possible, you would want to have as many threads as client connections and then set this number as high as possible


antonputra (approved):
Thank you! I'll definitely try this out.

PR #427, opened 2025-02-08
Go app HTTP and JSON alternatives

amitaid:
In the last few videos, you've compared Go's standard library with Rust, and have upgraded Rust's implementation to include libraries, though fairly standard (serde, tokio). I feel this comparison begins to be unfair in the other direction, as there are also Go community libraries that can add performance to that app.

I think a better comparison of both languages would be using the existing faster/more efficient alternatives that exist for both. In this PR I've included 3 alternative Go servers (Fiber, Gin and go/web), have replaced the JSON serializer with bytedance/sonic, added a yaml config param for server type.

These servers can probably be tuned better for performance, but I've included them mainly as suggestions for future tests. I also think Rust can be upgraded with frameworks like Actix, Axum and may_minhttp, but I'll leave it to someone who knows Rust better than me to add those.

I hope you find this useful, and use these in future comparisons.


antonputra (approved):
Thank you!

PR #428, opened 2025-02-11
Rust vs. Go (Golang): Performance (Fastest Frameworks + PostgreSQL)

antonputra:


vinhjaxt:
This is not a fair comparison, gnet does not fully support http, it is more like a network library, even if you parse the request with http standard lib, the way you build the response, it is still not practical

vinhjaxt:
Please refer to Techempower Benchmark

PR #429, opened 2025-02-11
Reduce the size of non-changing strings

LtdJorge:
This changes String values which don't change over time to Box<str>. String is 3 words in size while Box is 2, pretty trivial change, but it's the preferred style for production where every byte counts. If those strings needed to be shared, Rc<str> or Arc<str> would be better options with the same size but a tiny higher runtime cost.

I don't know if this would affect the performance in the benchmarks noticeably, but I'm sure it would reduce the memory size a bit if there are hundreds of thousands of requests in flight.

### DIFF
--- a/lessons/244/rust-app/src/device.rs
+++ b/lessons/244/rust-app/src/device.rs
@@ -3,6 +3,6 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Device {
     pub id: Option<i32>,
-    pub mac: String,
-    pub firmware: String,
+    pub mac: Box<str>,
+    pub firmware: Box<str>,
 }
--- a/lessons/244/rust-app/src/routes.rs
+++ b/lessons/244/rust-app/src/routes.rs
@@ -19,28 +19,28 @@ pub async fn get_devices() -> impl IntoResponse {
     let devices = [
         Device {
             id: Some(0),
-            mac: String::from("5F-33-CC-1F-43-82"),
-            firmware: String::from("2.1.6"),
+            mac: Box::from("5F-33-CC-1F-43-82"),
+            firmware: Box::from("2.1.6"),
         },
         Device {
             id: Some(1),
-            mac: String::from("44-39-34-5E-9C-F2"),
-            firmware: String::from("3.0.1"),
+            mac: Box::from("44-39-34-5E-9C-F2"),
+            firmware: Box::from("3.0.1"),
         },
         Device {
             id: Some(2),
-            mac: String::from("2B-6E-79-C7-22-1B"),
-            firmware: String::from("1.8.9"),
+            mac: Box::from("2B-6E-79-C7-22-1B"),
+            firmware: Box::from("1.8.9"),
         },
         Device {
             id: Some(3),
-            mac: String::from("06-0A-79-47-18-E1"),
-            firmware: String::from("4.0.9"),
+            mac: Box::from("06-0A-79-47-18-E1"),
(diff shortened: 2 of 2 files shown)


antonputra (approved):
Appreciate your feedback!

LtdJorge:
Forgot to mention that Box is only 2 words for Unsized types. A Box<&str> would be just one, but it would be a pointer to a pointer, the cool thing about Box is that it can wrap Unsized types as if it was a reference, but in owned fashion.

LtdJorge:
> actually u moved strings from stack to heap and it might slow down the calls because a lot of processor cycles waits instead of work

Not really, no. String is allocated on the heap. Of course, the variable where you assign the String is in the stack, but it's a Vec internally, so a pointer, a size and a capacity. And that pointer points to a heap allocation.

Anyway, in this case the strings are even 'static, so the allocations don't point to heap, they point to static memory, which is copied from the binary to a special read-only region in memory when the binary is executed.

snatvb:
> > actually u moved strings from stack to heap and it might slow down the calls because a lot of processor cycles waits instead of work
> 
> Not really, no. String is allocated on the heap. Of course, the variable where you assign the String is in the stack, but it's a Vec internally, so a pointer, a size and a capacity. And that pointer points to a heap allocation.
> 
> Anyway, in this case the strings are even 'static, so the allocations don't point to heap, they point to static memory, which is copied from the binary to a special read-only region in memory when the binary is executed.

Ah sorry. Strings everytime on heap. Yeah, with Box we can economy on size, you're right.
You reduced 8 byte per string as I get, 80 in total if I am not wrong

LtdJorge:
It's 16 bytes total for Box<str> and same for the Rc and Arc versions. String is 24. This is for 64bit machines, where a word (size of pointer, basically) is 8 bytes (64 bits). So the first variants are 2 words, String is 3. String is just a Vec internally, with custom UTF-8 checking in its methods.

If you want to quickly check the size of a type, you can just 
```rust
println!("Size of Box<str> {}", size_of::<Box<str>>());
```

A normal Box is only one word, but the standard library has specializations for many types. A pucture I like to check from time to time is https://www.reddit.com/r/rust/comments/74yrdp/rust_container_cheat_sheet_reposted/

snatvb:
> It's 16 bytes total for Box and same for the Rc and Arc versions. String is 24. This is for 64bit machines, where a word (size of pointer, basically) is 8 bytes (64 bits). So the first variants are 2 words, String is 3. String is just a Vec internally, with custom UTF-8 checking in its methods.
> 
> If you want to quickly check the size of a type, you can just
> 
> ```rust
> println!("Size of Box<str> {}", size_of::<Box<str>>());
> ```
> 
> A normal Box is only one word, but the standard library has specializations for many types. A pucture I like to check from time to time is https://www.reddit.com/r/rust/comments/74yrdp/rust_container_cheat_sheet_reposted/

Yaeh, as I said you saved 8 bytes - `24 - 16 = 8` and there is 10 strings that equals `80 bytes` in total :)

LtdJorge:
Yep, it's small, but with hundreds of thousands of requests per second, there might me many thousands of these at the same time. Anyway, the device structs here are pretty small, in the real world they'd be much bigger and it would have much more effect :D

snatvb:
> Yep, it's small, but with hundreds of thousands of requests per second, there might me many thousands of these at the same time. Anyway, the device structs here are pretty small, in the real world they'd be much bigger and it would have much more effect :D

I agree with you, not so small when we extrapolate on thousands requests per sec and I'm admiring that you can find optimism in even the smallest things, good job!

PR #431, opened 2025-02-11
Improve both apps for 244

SvetlinZarev:
### Proposed changes

1. Fix the processing of the COntent-Length header for the golang app. Without his change, all requests from JMeter were failing

2. Use `deadpool` instead of `bb8` as connection pool

3. Use `prepare_cached()` + `query_one()` instead of just `query_one()` for a significant speedup

After those changes, the rust app was significantly outperforming the golang one, which for some reason failed to scale beyond 64 DB connections.

JMeter tests:

* 20, 100 and 500 concurrent requests
* 8, 16, 24, 32 and 64 DB connections
* the apps * jmeter were running on different machines to avoid interference 

c8
C16
C24
C32
C64

### DIFF
--- a/lessons/244/go-app/cmd/gnet/server.go
+++ b/lessons/244/go-app/cmd/gnet/server.go
@@ -90,7 +90,7 @@ func (hc *httpCodec) parse(data []byte) (int, []byte, error) {
 }
 
 func (hc *httpCodec) getContentLength() int {
-	if hc.contentLength != -1 {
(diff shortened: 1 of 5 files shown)


SvetlinZarev:
PS: The bad results with golang-c64 are due to misconfiguration, so it can be ignored

antonputra (approved):
Thank you! I'll use it to compare Rust with Dragon

PR #433, opened 2025-02-12
Don't over-allocate in 244

richardpringle:
I'm not sure if this will make a difference or not, but since the Rust seems like it gets unstable at high cpu-usage, it might.

### DIFF
--- a/lessons/244/rust-app/Cargo.lock
+++ b/lessons/244/rust-app/Cargo.lock
@@ -695,8 +695,10 @@ name = "rust-app"
 version = "0.1.0"
 dependencies = [
  "axum",
+ "bytes",
  "deadpool-postgres",
  "serde",
+ "serde_json",
  "tokio",
  "tokio-postgres",
 ]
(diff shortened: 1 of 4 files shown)


SvetlinZarev:
IMO there is no point in over-complicating it. Personally I like to see benchmarks that are as close as possible to real world usage. There is no point in cheating like most people do in techempower. 

If you want to reduce allocations either use `Cow` and get rid of the custom `Json<T>` impl. It's just unnecessary complexity

antonputra (approved):
Thank you for the PR, I'll compare your implementation as well, just to see if there is any difference.

LtdJorge:
I originally chose to go with Box<str> because in a real use case, you would not have static strings for the addresses, you’d probably get them from the DB or API or something like that. So there would have to be an allocation at some point.

If there was only one device retrieved and that was the only one served, then there would only be one allocation, and many ‘static references to it. However, by returning the same static string every time, there zero allocations, and I had that idea at first but discarded it because it seemed more like a “drag racing” trick for benchmarking than a real thing 😅

SvetlinZarev:
I agree. 

On the other hand, the golang app does not allocate new strings on each request and works as if using `&'static str` in rust.

I wonder how much the golang perf would suffer if it was allocating

Issue #430, opened 2025-02-11
use ntex or  may-minihttp or xitca-web all scorde the highest score in tech empowered benchmarks


driver005:
here is the link to the results https://www.techempower.com/benchmarks/#hw=ph&test=composite&section=data-r22&l=yw2l8f-cn3

ntex comes first followed by may-minihttp and xitca-web

They are way faster than axum, hyper or axtix from the look of the Composite score

SvetlinZarev:
I', quite interested by ntex. It's author is the original author of ActixWeb, so I expect it to be pretty fast

antonputra:
@SvetlinZarev would you be able to contribute to and submit the ntex implementation?

antonputra:
I can only do 3 frameworks at a time, otherwise, it becomes very expensive. Maybe May vs. Ntex vs. Axum.

SvetlinZarev:
@antonputra Sure. The same route as in the rust-app (axum) ? With or without DB ?

antonputra:
@SvetlinZarev with database, let me post here as well

1. Test
`curl -i http://localhost:8080/api/devices`

```bash
HTTP/1.1 200 OK
Server: gnet
Date: Thu, 13 Feb 2025 01:21:35 GMT
Content-Length: 271
Content-Type: application/json

[{"id":0,"mac":"5F-33-CC-1F-43-82","firmware":"2.1.6"},{"id":1,"mac":"44-39-34-5E-9C-F2","firmware":"3.0.1"},{"id":2,"mac":"2B-6E-79-C7-22-1B","firmware":"1.8.9"},{"id":3,"mac":"06-0A-79-47-18-E1","firmware":"4.0.9"},{"id":4,"mac":"68-32-8F-00-B6-F4","firmware":"5.0.0"}]
```

2. Test (Postgres)

`curl -i http://localhost:8080/api/devices -d '{"mac": "81-6E-79-DA-5A-B2", "firmware": "4.0.2"}' -H "Content-Type: application/json"`

```bash
HTTP/1.1 201 Created
Server: gnet
Date: Thu, 13 Feb 2025 01:23:00 GMT
Content-Length: 57
Content-Type: application/json

{"id":66009,"mac":"81-6E-79-DA-5A-B2","firmware":"4.0.2"}
```

DB Schema/migration*

```sql
  --
  -- Create application users.
  --
[...]

PR #434, opened 2025-02-13
Implement rust app based on May

KonstantinKuklin:
a part of #430


antonputra (approved):
Thank you! I'll test it soon, maybe by the end of this week.

antonputra:
@KonstantinKuklin, sorry it took me a while. Would you be able to update this code, and I'll run it in a couple of days?

PR #442, opened 2025-03-04
Nginx vs. HAProxy vs. Traefik: Performance

antonputra:


yonathan9669:
Hey man, awesome job 🙏!!

I was wondering if you have considered configuring the VMs (starting with the proxies) using `cgroups` to help stabilize the services during peak times. The goal would be to prevent `bufferbloat` behavior by allowing enough resources for the machine to handle the rest of the environment properly. Based on your charts, setting the CPU usage to around `80%` could be a good starting point.

I'm eager to see if this helps you. 👍🏼

antonputra:
@yonathan9669 I didn't get it. Cgroups are usually used to isolate different processes from each other running on the same server. I run my proxies on different VMs, how can configuring cgroups in this case help?

yonathan9669:
Hey @antonputra,

`Cgroups` can indeed be used to isolate processes and limit system resources. While tools like `cpulimit` or `systemd` can achieve similar results, I've had particularly good experiences improving resource usage with `cgroups`.

In my humble opinion, using `cgroups` might help because it appears that your services are experiencing a `bufferbloat`-like issue. This means that the CPU –in this case– becomes a bottleneck as it doesn't have enough "free quota" to properly handle networking routing and other processes.

Looking at your charts reminds me of past struggles with similar issues when deploying local servers. Limiting resource usage close to the peaks while leaving enough "free space" for I/O can help stabilize performance and improve the overall appearance of your charts.

If you need any help, I'd be more than happy to support you in preparing the environments to run some tests.

Additionally, I found some content that extends what I'm trying to express: [Linux Beyond the Basics: CPU Management with CPU and Cpuset Cgroups](https://medium.com/@weidagang/linux-beyond-the-basics-cpu-management-with-cpu-and-cpuset-cgroups-a4853bc645f5)."

antonputra:
@yonathan9669 Thank you for the info. I had never heard about this before but will definitely take a deeper look.

PR #444, opened 2025-03-05
249 TCP vs UDP: optimize

The-Emperor10:
This optimizes the TCP vs UDP benchmark. I don't know how to use prometheus so I haven't benchmarked it myself but TCP will likely have a minor improvement while UDP throughput should increase greatly.

The largest optimization is batching several requests together into one call of `send`/`recv`, which will cause UDP to only send 1 packet for several requests and should reduce the number of syscalls and context switches for both tests. The kernel will already batch the TCP requests together into one packet before its sent out before this change which is likely the primary cause for the massive throughput difference between TCP and UDP.

There are other minor changes that will probably marginally increase performance for both tests and I fixed all the compiler warnings I was getting with `-Wall -Wextra -Wpedantic` on gcc. Note that the `UDP_BUFSIZE` define in `monitoring.hpp` can be changed to be much larger when testing on `localhost` vs over a network.

### DIFF
--- a/lessons/249/app/tcp_client.cpp
+++ b/lessons/249/app/tcp_client.cpp
@@ -8,7 +8,6 @@
 #include "monitoring.hpp"
 
 #define BACKLOG 20
-#define MAXDATASIZE 90
 #define PORT "8080"
 // #define ADDRESS "tcp-server.antonputra.pvt"
 #define ADDRESS "localhost"
@@ -18,8 +17,7 @@ using namespace prometheus;
 int main()
 {
     const char *enable_monitoring_env = std::getenv("ENABLE_MONITORING");
-
-    std::string enable_monitoring(enable_monitoring_env);
+    bool enable_monitoring = enable_monitoring_env && std::strcmp(enable_monitoring_env, "true") == 0;
 
     // Expose Prometheus metrics on port 9081.
     Exposer exposer{"0.0.0.0:9080"};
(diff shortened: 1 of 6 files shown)


antonputra (approved):
Thank you so much! I'll rerun the test in a couple of days.

lessons/249/app/udp_client.cpp
```diff
@@ -47,13 +42,31 @@ int main()
     hints.ai_flags = AI_PASSIVE;
 
     // Fill out the struct. Assign the address of of the server to the socket structure, etc.
-    getaddrinfo(ADDRESS, PORT, &hints, &servinfo);
+    int err;
+    if ((err = getaddrinfo(NULL, PORT, &hints, &servinfo)) != 0)
```
antonputra, line 46: @The-Emperor10 Is it a copy-paste mistake? It may work on localhost, but we still need to use ADDRESS to resolve the remote host, right?

lessons/249/app/tcp_client.cpp
```diff
@@ -66,7 +61,12 @@ int main()
     hints.ai_flags = AI_PASSIVE;
 
     // Fill out the struct. Assign the address of of the server to the socket structure, etc.
-    getaddrinfo(ADDRESS, PORT, &hints, &servinfo);
+    int err;
+    if ((err = getaddrinfo(NULL, PORT, &hints, &servinfo)) != 0)
```
antonputra, line 65: @The-Emperor10 same here

lessons/249/app/udp_client.cpp
The-Emperor10: Yeah copy paste mistake

PR #445, opened 2025-03-08
Lesson 250: Update nginx.conf to remove gzip compression to better compare to haproxy

gheffern:


gheffern:
It appears haproxy is not configured to do compression in lesson 250 but nginx is. I removed the gzip config for nginx to enable a better comparison.

antonputra (approved):
Sure, makes sense

PR #449, opened 2025-03-14
Elixir: use built-in json + fix datetime insertion

AstolfoKawaii:
This pull request consists of two changes:

1. Switch from Jason to the new standard JSON library
2. Convert created_at and updated_at fields to string before insertion

### Switch from Jason to the new standard JSON library

Elixir 1.18 introduced built-in support for JSON, which according to my tests is significantly faster than Jason

```
##### With input devices_struct #####
Name                                ips        average  deviation         median         99th %
JSON.encode_to_iodata!/1       928.40 K        1.08 μs  ±4567.52%        0.93 μs        1.48 μs
Jason.encode_to_iodata!/1      494.06 K        2.02 μs  ±1756.63%        1.76 μs        3.28 μs
Comparison:
JSON.encode_to_iodata!/1       928.40 K
Jason.encode_to_iodata!/1      494.06 K - 1.88x slower +0.95 μs

##### With input devices_json #####
Name                      ips        average  deviation         median         99th %
JSON.decode!/1       328.70 K        3.04 μs   ±888.06%        2.81 μs        4.58 μs
Jason.decode!/1      235.01 K        4.26 μs   ±676.94%        3.59 μs        6.63 μs
Comparison:
JSON.decode!/1       328.70 K
Jason.decode!/1      235.01 K - 1.40x slower +1.21 μs
```

### Convert created_at and updated_at fields to string before insertion

Apparently elixir version of the app **never inserted anything into the database**.
Without `DateTime.to_string()` postgrex returns an error similar to:
`{"message":"Postgrex expected a binary, got ~U[2025-03-14 14:00:56.233724Z]. Please make sure the value you are passing matches the definition in your table or in your query or convert the value accordingly."}` with _400 Bad Request_ status code

It was present here: 6d63c9e0382f422ec6fe93d7f8e9b837ea553205
But is missing here: f991ece4c02c2963b5b9746d52d002646922d885 (the commit that you used for the benchmark)

---

\+ updated dependencies as was [recommended](https://github.com/antonputra/tutorials/pull/370#issuecomment-2561268019) by @josevalim


josevalim:
Thank you @AstolfoKawaii! Good catch on the DateTime string conversion.

FWIW, I ran the benchmarks locally with and without the `to_string` conversion, and the numbers are roughly the same on my machine but maybe the servers will tell a different story. On the other hand, I'd have expected the _benchmarks to have failed_ if the server was returning 400, as some basic smoke tests should be in place.

If someone runs the benchmarks again, I'd love if the benchmarks numbers and server logs are made available for posterity. The benchmark process has been opaque: improvements we see locally are not seen in the results and the benchmarks show failures we cannot reproduce locally. Having a bit more transparency would likely help everyone: it could provide us feedback and help spot mistakes like the one in this issue.

AstolfoKawaii:
> The benchmark process has been opaque: improvements we see locally are not seen in the results and the benchmarks show failures we cannot reproduce locally.

I agree. It seems like the version used for the benchmark was somehow different from the one in the repo, because the one in the repo shouldn't have worked at all 😅

antonputra (approved):
Thank you! Next time when i run this bench I'll do it on VMs.

yordis:
We are observing huge span gaps in our traces, up to entire second or more sometimes; and most people, including myself, feels that it is due to Kubernetes CPU limiting.

Ideally the throttling would be 0 so when those high CPU usage happen it doesn't affect the system as much and let Erlang perform at 100%

@josevalim I wonder if that is the precise issue that Anton is facing in his tests. I can not find any other explanation for his and our situation by now. I used the following functions to verify, at the very least, the allocated resources:

```elixir
:erlang.system_info(:cpu_quota)
:erlang.system_info(:schedulers)
:erlang.system_info(:schedulers_online)
```

@antonputra definitely curious to see how much VMs without k8s help here, but even more curious if you remove the CPU Limiting.

AstolfoKawaii:
> We are observing huge span gaps in our traces

I apologize, but could you clarify which traces you are referring to? :sweat_smile:

antonputra:
@yordis Sorry for the delay. Over the last year or so, while running these benchmarks, I found that different applications, and especially configurations associated with deployment, are affected differently by Kubernetes. Once in a while, I do run my tests on VMs, but I should probably even switch to bare metal EC2s.

yordis:
@antonputra we had to add Node Poll to isolate the pods, and run one pod per Node without any CPU Limits to stabilize the situation.

That could be the compromise, still k8s but treat the situation as if you are giving the entire VM to one elixir app.

The bottom line is, we can not interfere with Erlang scheduler.

Lastly, we removed the `livenessProbe` all together because it was causing issues:

> 17 Unhealthy: Liveness probe failed: command "/app/myapp/bin/myapp pid" timed out

So worth removing the `livenessProbe` and just observe the container exiting (which is the default behavior I believe)

realhaidinh:
Have you tried tuning the Erlang VM? I think it might help
https://docs.pleroma.social/backend/configuration/optimizing_beam/
https://www.rabbitmq.com/docs/next/runtime#vm-settings

PR #458, opened 2025-04-15
chore: improve k8s deployment

yordis:


antonputra:
why not?

yordis:
> Why not? 🤷🏻‍♂️

We've all tried to make it work. In short, it is what it is. The most reliable and performant way to run an Erlang application is by applying the prescribed best practices. I will invite you to join Elixir Slack server and ask in `#opentelemetry` channel, they are a plenty of experienced people.


Recently, we encountered this exact issue. We had to remove resource limits, add a node group to avoid noisy neighbors, and isolate workloads to fully utilize resources without causing significant outages.

In your video, you might have noticed a lot of CPU throttling, which worsens the performance of the Erlang scheduler. Erlang requires the ability to spike to 100% CPU usage to continue task execution. While this might increase latency, the system will remain operational. You don't want anything interfering with the Erlang scheduler's functioning. Otherwise it makes everything worst, causing even more timeouts.

When such throttling occurs, you definitely do not want the livenessProbe to shut down the container, as that could exacerbate the issue by reallocating workloads to other containers, potentially leading to cascading failures.

Some people suggest you to run Erlang/Elixir in a VM instead of k8s, that is the alternative to avoid all these misconfigurations. But the intent is the same, let Erlang to control as much as the resources as possible.

A good reference to have around: https://adoptingerlang.org/docs/production/kubernetes/

antonputra (approved):

PR #464, opened 2025-05-01
Use postgres from deno.land/x/postgresjs

PanovYury:
Deno currently imports postgresjs directly from npm. However, there is a dedicated Deno-compatible version available at https://deno.land/x/postgresjs@v3.4.5.

The postgresjs project itself includes Deno-specific code within the deno directory (see: https://github.com/porsager/postgres/tree/master/deno). This suggests that using the Deno-specific version could offer performance benefits or other advantages.

I believe utilizing the Deno-compatible version of postgresjs could be an optimization.


antonputra (approved):
Thanks! It makes sense.

PR #465, opened 2025-05-05
Folder raname for easier navigation 

domik82:
Hi,

Let me know if you would accept such change in your repo. If not just reject the PR. I used link in the Readme and extracted the title of the video so that folders have same names. It's easier to search for such folders and it seems AI tooling I use to learn from this repo is able to figure out stuff faster. 

If for some reason that doesn't fit you that is fine. 

One readme pointed to video that was having wrong link but video actually existed. 

I fixed all docs/contents.md to point to proper folders.


antonputra:
I get your point, but I use short folder names primarily so that they take up less space in the terminal while I record tutorials.
Screenshot 2025-05-06 at 10 01 25 PM

PR #474, opened 2025-09-05
Example of coroutines with boost

grincha09:
Thanks for your videos
Just wanted to share my approach of using boost asio with coroutines 
Coroutines are stackfull and also from boost (there are also stackless from cpp-20 - the channel "Talking Async" on youtube  shows examples of use) 
No idea how it affects performance, but it is more readable (at least for me) and does not use explicit shared pointers(which do worsen performance)
Hope you find something useful


antonputra:
thnk you for the pr, i'll give it a try

antonputra (approved):
Thank you!

PR #494, opened 2025-12-14
fix: correct helm provider kubernetes block syntax

lakmalniranga:
This PR updates the helm provider configuration to be compatible with v3.0.0, which has introduced breaking changes to the provider syntax.

Reference: https://github.com/hashicorp/terraform-provider-helm/pull/1559/commits/5a43269fe9c37d4824171ddd44bd2d44347fbe89

### DIFF
--- a/lessons/196/terraform/11-helm-provider.tf
+++ b/lessons/196/terraform/11-helm-provider.tf
@@ -7,7 +7,7 @@ data "aws_eks_cluster_auth" "eks" {
 }
 
 provider "helm" {
-  kubernetes {
+  kubernetes = {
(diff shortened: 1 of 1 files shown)


lakmalniranga:
Closed: I later realized I was already using the latest version. It also requires a minor syntax change for set as well.

For example:

```
  set = [
    {
      name  = "clusterName"
      value = aws_eks_cluster.eks.name
    },
    {
      name  = "serviceAccount.name"
      value = "aws-load-balancer-controller"
    },
    {
      name  = "vpcId"
      value = aws_vpc.main.id
    }
  ]
```

PR #498, opened 2025-12-20
275 bun optimized

solmanter:
#### Intro
Hi Anton! I watched your Go and Typescript comparison video (the latest one) and I want to improve on this code. I just updated bun to be more optimal. Thanks for your work!

#### Changes
- Simplified user API routes in app.ts, ensuring proper response status codes.
- Updated database connection string format in db.ts for PostgreSQL.
- Enhanced user saving logic in users.ts to directly use the User object.
- Changed config loading method in config.ts to use Bun's file API for better compatibility.
- Removed error logging to be fair with go

### DIFF
--- a/lessons/275/bun-app/package.json
+++ b/lessons/275/bun-app/package.json
@@ -11,7 +11,7 @@
     "typescript": "^5"
   },
   "scripts": {
-    "build": "bun build --compile --minify --sourcemap ./src/app.ts --outfile bun-app"
+    "build": "bun build --compile --minify --sourcemap --sql-preconnect ./src/app.ts --outfile bun-app"
   },
   "dependencies": {
     "js-yaml": "^4.1.1"
--- a/lessons/275/bun-app/src/app.ts
+++ b/lessons/275/bun-app/src/app.ts
@@ -1,67 +1,51 @@
 import config from "./config.ts";
 import { save } from "./users.ts";
 import type { User } from "./types.ts";
-import type { Config } from "./config.ts";
-
 
 const server = Bun.serve({
   // Timeout in seconds to match Node.js
   idleTimeout: 60,
   development: false,
   reusePort: true,
-  port: (config as Config).appPort,
+  port: config.appPort,
   async fetch(req: Request): Promise<Response> {
     const path = new URL(req.url).pathname;
-    if (path === "/healthz") return new Response("OK");
 
-    if (req.method === "GET" && path === "/api/users") {
-      const users: User[] = [
-        {
-          id: 1,
-          name: "David D. Patton",
-          address: "1670 Stiles Street",
-          phone: "412-578-3857",
-          image: `user.png`,
-        },
-        {
-          id: 2,
-          name: "Gary E. Eaton",
-          address: "828 Collins Avenue",
-          phone: "614-866-1660",
-          image: `user.png`,
-        },
-        {
-          id: 3,
-          name: "John J. Fox",
-          address: "1895 Columbia Mine Road",
-          phone: "304-505-3622",
-          image: `user.png`,
-        },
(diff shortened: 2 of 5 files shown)


antonputra (approved):
Thanks! I really appreciate your help. I'll definitely use it for my next benchmarks!

lessons/275/bun-app/package.json
```diff
@@ -11,7 +11,7 @@
     "typescript": "^5"
   },
   "scripts": {
-    "build": "bun build --compile --minify --sourcemap ./src/app.ts --outfile bun-app"
+    "build": "bun build --compile --minify --sourcemap --sql-preconnect ./src/app.ts --outfile bun-app"
```
RiskyMH: `--sql-preconnect` is for runtime, not `bun build`

lessons/275/bun-app/src/db.ts
```diff
@@ -1,10 +1,14 @@
 import config from "./config.ts";
 import { SQL } from "bun";
-import type { Config } from "./types.ts";
+
+const { db } = config;
 
 const sql = new SQL({
-  url: `postgres://${(config as Config).db.user}:${(config as Config).db.password}@${(config as Config).db.host}:5432/${(config as Config).db.database}`,
-  max: (config as Config).db.maxConnections,
+  url: `postgresql://${db.user}:${db.password}@${db.host}:5432/${db.database}`,
```
RiskyMH: I haven't measured perf difference, but you don't need to construct the url here as there are options for more individual instead

https://bun.com/docs/runtime/sql#postgresql-options

RiskyMH (commented):
Just some comments on how one could improve this further. I can make a pr if you'd like.

antonputra:
@RiskyMH Thank you for reviewing this!

lessons/275/bun-app/src/app.ts
```diff
@@ -1,67 +1,51 @@
...
-          id: 3,
-          name: "John J. Fox",
-          address: "1895 Columbia Mine Road",
-          phone: "304-505-3622",
-          image: `user.png`,
-        },
-      ];
-      return Response.json(users);
-    }
+    if (path === "/healthz") return new Response("OK", { status: 200 });
```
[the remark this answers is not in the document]
solmanter: Static routing is indeed faster, but we keep it around because we aim to simulate real-world scenarios as closely as possible. and similar code structure like go

lessons/275/bun-app/src/app.ts
```diff
@@ -1,67 +1,51 @@
...
-      const user: User = {
-        ...incomingUser,
-        createdAt: datetime,
-        updatedAt: datetime,
-        image: key,
-      };
+      if (req.method === "POST") {
+        const incomingUser = await req.json() as User;
+        const now = Date.now();
+        const datetime = new Date(now);
```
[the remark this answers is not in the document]
solmanter: yeah, could be

lessons/275/bun-app/package.json
solmanter: your are right, that should on runtime

lessons/275/bun-app/src/db.ts
solmanter: yes, there are not difference, i just updated types and made more readable

lessons/275/bun-app/package.json
solmanter: That is interesting to me, i can't find the reason in my tests and benchmarks. It always stayed the same

RiskyMH: All this pre connect does is start the connection before javascript code. Which means in theory faster... but as its not set in DATABASE_URL, this flag can't be used

PR #500, opened 2025-12-30
refactor: better python performance

Lewiscowles1986:
So I tried a few things, which get the minimum time in k6 down to microseconds for a request. They set slightly more aggressive timeouts, use pydantic to marshal the response from the POST endpoint to the DB. ~and set a latency on every request DB calls of 1 second.~ (the set local DB call, actually creates some overhead)

Ive not re-run the entire test, instead I've contributed a k6 test, which I used to verify on my local machine; within the also modified docker-compose yaml.

The migrations are still manual; and you'll probably notice I tried increasing server connections. I Think adjusting the `min_size` of db connections could be another good way to gain performance.

I'm pretty sure it still wont top bun, but you could get a few more thousand requests per second

### DIFF
--- a/lessons/276/fastapi-app-simpler/Dockerfile
+++ b/lessons/276/fastapi-app-simpler/Dockerfile
@@ -22,7 +22,7 @@ RUN addgroup --system app && adduser --system --ingroup app app
 
 COPY --from=build /usr/local/lib/python3.14/site-packages /usr/local/lib/python3.14/site-packages
 
-COPY --from=build /usr/local/bin/gunicorn /usr/local/bin/gunicorn
+COPY --from=build /usr/local/bin/granian /usr/local/bin/granian
 
 COPY ./src /app
 COPY ./config.yaml /app/config.yaml
@@ -34,4 +34,4 @@ ENV MALLOC_ARENA_MAX=2
 
 USER app
 
-CMD ["gunicorn", "-w", "1", "-k", "uvicorn.workers.UvicornWorker", "--preload", "--timeout", "10", "--graceful-timeout", "5", "--log-level", "error", "main:app", "--bind", "0.0.0.0:8080"]
+CMD ["granian", "--port", "8080", "--interface", "asgi", "--host", "0.0.0.0", "--process-name", "fastapi speedrun", "main:app"]
--- a/lessons/276/fastapi-app-simpler/requirements.txt
+++ b/lessons/276/fastapi-app-simpler/requirements.txt
@@ -11,7 +11,7 @@ fastapi[standard]==0.127.1
 fastapi-cli==0.0.20
 fastapi-cloud-cli==0.8.0
 fastar==0.8.0
-gunicorn==23.0.0
+granian[pname]==2.6.0
 h11==0.16.0
 httpcore==1.0.9
 httptools==0.7.1
@@ -45,7 +45,6 @@ typing-inspection==0.4.2
 typing_extensions==4.15.0
 urllib3==2.6.2
 uuid6==2025.0.1
-uvicorn==0.40.0
 uvloop==0.22.1
 watchfiles==1.1.1
 websockets==15.0.1


Lewiscowles1986:
Maybe @emerson-proenca; If you make a PR  I'd love to run it. I've not had a lot of experience with it, so I consider it to be esoteric at the moment; but maybe it's very easy and I'm just unclear about approach.

I do know AI was telling me I was barking up the wrong tree with this, but then when I looked at all my latencies, they seemed to support me, more than AI review of my work.

Lewiscowles1986:
Here are some logs from K6 test included with PR, run on an M2 Air laptop

```
    scenarios: (100.00%) 1 scenario, 100 max VUs, 1m0s max duration (incl. graceful stop):
              * default: 100 looping VUs for 30s (gracefulStop: 30s)

--- some warnings

     ✗ is status 201
      ↳  99% — ✓ 79347 / ✗ 489

     checks.........................: 99.38% ✓ 79347     ✗ 489  
     data_received..................: 25 MB  821 kB/s
     data_sent......................: 15 MB  509 kB/s
     http_req_blocked...............: avg=10.61µs min=0s       med=1µs     max=13.98ms p(90)=2µs     p(95)=3µs    
     http_req_connecting............: avg=8.42µs  min=0s       med=0s      max=12.88ms p(90)=0s      p(95)=0s     
     http_req_duration..............: avg=37.32ms min=16µs     med=21.07ms max=1.8s    p(90)=40.88ms p(95)=78.71ms
       { expected_response:true }...: avg=37.52ms min=3.7ms    med=21.11ms max=1.8s    p(90)=41.02ms p(95)=79.22ms
     http_req_failed................: 0.61%  ✓ 489       ✗ 79347
     http_req_receiving.............: avg=21.23µs min=0s       med=12µs    max=5.59ms  p(90)=32µs    p(95)=43µs   
     http_req_sending...............: avg=6.73µs  min=1µs      med=4µs     max=4.93ms  p(90)=10µs    p(95)=15µs   
     http_req_tls_handshaking.......: avg=0s      min=0s       med=0s      max=0s      p(90)=0s      p(95)=0s     
     http_req_waiting...............: avg=37.29ms min=12µs     med=21.04ms max=1.8s    p(90)=40.83ms p(95)=78.69ms
     http_reqs......................: 79836  2660.1558/s
     iteration_duration.............: avg=37.57ms min=159.87µs med=21.28ms max=1.8s    p(90)=41.12ms p(95)=78.98ms
     iterations.....................: 79836  2660.1558/s
     vus............................: 100    min=100     max=100
     vus_max........................: 100    min=100     max=100


running (0m30.0s), 000/100 VUs, 79836 complete and 0 interrupted iterations
default ✓ [======================================] 100 VUs  30s

```

Lewiscowles1986:
Apologies, I realised after re-running the base-line I'd committed something I should not have. I've commented it out, but basically, I doubled the number of queries being run as part of a failed experiment and then accidentally `git add .` it. The rest is done now

antonputra (changes requested):
Thank you, @Lewiscowles1986, for the PR. Sorry I don't have clear guidelines for pull requests. Would you be able to just create a fastapi-app-v2 folder and keep the original as is? It's just easier to compare in the future.

Lewiscowles1986:
To be honest, it merging isnt' the most important thing as you'll only get at most a few thousand requests (not per second, overall 😄 )

Thank you for the tutorial; if you'd still like the changes I'll make them

antonputra:
@Lewiscowles1986 Yes please, I'll keep it. When it's time to refresh the benchmark, I'll go over and test everything again.

Lewiscowles1986:
It now points to a separate folder; sorry I've been unwell; it's also simpler than the original patch I'd sent over

Spitfire1900:
Many thanks to everyone!  I would recommend setting workers explicitly for both grainian and bun to ensure better fairness in benchmark; I noticed guvicron was set to 1 worker in the original test.


It's always important to have a benchmark that has "average code" rather than performing esoteric operations or configuration.

PR #502, opened 2025-12-31
Replace Gunicorn+uvicorn with Granian for FastAPI

ondrados:
Swapped out gunicorn + uvicorn for [granian](https://github.com/emmett-framework/granian) (Rust-based ASGI server). Since we're mostly doing I/O, the Rust runtime helps. Saw decent gains in local wrk tests.


antonputra (changes requested):
Thanks @ondrados for the PR. Would you mind creating `Dockerfile.granian` instead and keeping the original as is?

antonputra (approved):
Thanks!

PR #518, opened 2026-04-02
Update to .net 10, house both project (non aot and aot) under one solution + update nuget packages

Seabizkit:
1.  So i moved the 2 folders under a sub folder so i could bring the solution file out.. a level
  so that i could have 1 solution file which allows me to view both projects at the same time, as that allows for starting to share base resources later.. Makes no sense to maintain separately. 
  
2. updated to .net 10
  
3. updated Nuget Package referances.

4. updated the docker files to target 10.
     I do not know if what was there was working, but the images do exist.

5. Hoping for a new Video

PS im no goland expert but go is doing 

imgKey := fmt.Sprintf("go-thumbnail-%d.png", counter)

while c# is doing 

 var id = Interlocked.Increment(ref StaticData.Counter) - 1;
 var image = new Image($"cs-thumbnail-{id}.png");

these are not the samething, shouldnt it be something like

```
Go atomic version
atomic.AddInt64(&counter, 1)
C# version
Interlocked.Increment(ref counter)
```

PSS - i think you wnat to change Device to a struct

```
public struct Device
{
    public string Uuid;
    public string Mac;
    public string Firmware;
}

public static readonly Device[] Devices =
[
    new() { Uuid = "...", Mac = "...", Firmware = "2.1.6" },
    new() { Uuid = "...", Mac = "...", Firmware = "2.1.5" },
    new() { Uuid = "...", Mac = "...", Firmware = "3.0.0" },
    new() { Uuid = "...", Mac = "...", Firmware = "1.0.1" },
    new() { Uuid = "...", Mac = "...", Firmware = "3.5.6" }
];
```


Seabizkit:
@HaikAsatryan could you review, so we can hopefully get repo owner to merge. Also didn't want to go too far.. But this simplifies maintenance

HaikAsatryan:
@Seabizkit 
Quick disclaimer first: this repo isn't mine, @antonputra is the owner. Just dropping a friendly review since I did the .NET 9 upgrade a while back.

Overall the structural move and version bumps look good. A few things worth addressing before merge:

AOT Dockerfile (cs/cs-app-aot/Dockerfile) has a multi-arch bug. Line 8 passes both -a $TARGETARCH and -r linux-x64. The -r linux-x64 overrides the arch, so an arm64 build will silently produce an x64 binary. Drop -r linux-x64 and let -a $TARGETARCH handle the RID.

AOT Dockerfile is also missing USER $APP_UID that the JIT Dockerfile has. Worth keeping them consistent.

build.sh wasn't updated for the folder move. With APP_DIR=cs/cs-app, the resulting image tag contains a slash (aputra/cs/cs-app-202-arm64), which Docker Hub rejects. Either sanitize the tag in the script or flatten the folder names.

Packages: Npgsql 10.0.2 and prometheus-net 8.2.1 are current. AWSSDK.S3 is at 4.0.20.2; latest is 4.0.23 if you want to bump.

On the Device struct suggestion: I'd keep it a class. Devices is a static readonly Device[] of 5 entries initialized once at startup, so there's no allocation pressure to optimize away, and the class keeps required + init semantics.

On the counter mismatch: you're right that the two aren't equivalent, but I'd argue the fix belongs on the Go side. I'm not a Go expert, but counter += 1 in Go will race under concurrent requests and undercount, so the C# Interlocked.Increment is the correct behavior and Go should use atomic.AddInt64 to match. There are a few other Go/.NET parity issues worth a separate pass too (Go recreates the S3 client per request, formats SQL with fmt.Sprintf per call, uses the now-EOL aws-sdk-go v1, and uses pgx defaults with no prepared statements vs Npgsql's multiplexing + auto-prepare), but those are outside the scope of this PR.

